PR #2, opened 2017-01-30
Install kubectl in gcloud builder

viglesiasce:


skelterjohn:
Thanks, and sorry that it took so long to notice this PR.

Unfortunately we cannot accept PRs to this repo - it is currently "mirror-out" from our internal codebase. I'll add it from the inside.

skelterjohn:
Now that I think about usage a bit, I'm not so sure that this PR is sufficient. `kubectl` also needs to fetch IP/user/pass for the cluster, and store that on disk to be used. The only fs that persists from step to step is in /workspace and that's not a good place to put things. I'll think about the best way to integrate this into our toolbox and update this issue when we've got something.

viglesiasce:
Sounds good @skelterjohn. I was going to go about it by doing something like this for the command:

```
gcloud container clusters get-credentials blah && kubectl apply -f deployment.yaml
```

skelterjohn:
That would work, certainly, but doesn't sit well with me for usage I want to document or put in examples :)

Maybe a kubectl builder that uses the gcloud builder as a base image, and has a wrapper script that checks some env vars? In this case, we could piggy-back on the env already used by gcloud and have people set CLOUDSDK_CONTAINER_CLUSTER=foobar, and the wrapper script does as you indicate.

viglesiasce:
<opinion>The UX of having it all spelled out and obvious in the command seems much better than a tucked in interpretation of an env variable with a wrapper script.</opinion>

skelterjohn:
Thanks, I appreciate your thoughts there. I've got something that allows either to work - you can spell it out like you do in your example (if you use sh as the entrypoint), or you can provide the env vars necessary.

Unfortunately, this is blocked on the scopes we give to your build. Soon-ish we hope to give your builds the https://www.googleapis.com/auth/cloud-platform scope (ie, everything) and rely on the robot permissions for restrictions. But right now the scopes don't allow getting credentials for the cluster.

When this changes, I'll update this thread.

Regner:
Don't suppose any progress has been made on this? Would really love to be able to deploy more easily to my GKE cluster from Cloud Builder.

DocBradfordSoftware:
> Would really love to be able to deploy more easily to my GKE cluster from Cloud Builder.
Did you see this discussion. #22

Regner:
I did not so thanks for linking that. :)

iMelnik:
As there is a workaround to get kubectl working in #22, I guess, there is nothing wrong with having kubectl installed inside the image. WDYT?

skelterjohn:
It's certainly possible for people to use kubectl today, with some stretching. But for a version we officially support, it's important that it work well out of the box. Hopefully resolved soon, stay tuned!

bendory:
Closing; we now have a supported [kubectl Build Step](https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/kubectl). Please re-open (or open a new PR or issue) if our supported `kubectl` does not meet your needs.

(Note: since @skelterjohn's original reply on Feb 21, we are now able to accept PRs.)

PR #6, opened 2017-03-07
Fixing example link in glide README

domudall:
Most other example directories use `_` separators, so not sure if it's the link that should be changed or the directory.


skelterjohn:
Thanks! Unfortunately our current setup does not allow PRs to this repo (we'll change this soon). But I'll fix the issue from inside.

domudall:
Ah OK, thanks!

PR #7, opened 2017-03-09
documentation fix for npm example

vorozhko:


Philmod:
Thanks, I fixed the doc internally and it will be propagated to github soon.

PR #8, opened 2017-03-09
documentation fix for npm example

vorozhko:


skier31415:
Thank you, updated internally, will be mirrored soon.

PR #10, opened 2017-03-10
add docker-compose

skippy:
I think this would be quite useful to have for those of us that want to:
* build
* run tests against that build
* ship that same build

### DIFF
--- a/docker-compose/Dockerfile
+++ b/docker-compose/Dockerfile
@@ -1,0 +1,11 @@
+FROM ubuntu:trusty
+
+# Based on instructions from:
+# https://docs.docker.com/compose/install/
+RUN \
+   apt-get -y update && \
+   apt-get -y install ca-certificates curl && \
+   curl -L "https://github.com/docker/compose/releases/download/1.11.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose && \
+   chmod +x /usr/local/bin/docker-compose
+
+ENTRYPOINT ["/usr/local/bin/docker-compose"]
--- a/docker-compose/README.md
+++ b/docker-compose/README.md
@@ -1,0 +1,31 @@
+# docker-compose
+
+This is a tool builder to simply invoke
+[`docker-compose`](https://docs.docker.com/compose/) commands.
+
+Arguments passed to this builder will be passed to `docker-compose` directly, allowing
+callers to run [any docker-compose
+command](https://docs.docker.com/compose/reference/overview/).
+
+By default, the version of docker-compose that is used by this builder is `1.11.2`.
+
+## Examples
+
+The following examples demonstrate build requests that use this builder:
+
+### Build and push a container image
+
+This `cloudbuild.yaml` invokes a `docker-compose run`, followed by a build, and then
+pushes the resulting image.
+
+```
+steps:
+- name: gcr.io/cloud-builders/docker-compose
+  args: ['run', 'app', 'test.sh']
+  id:   'test'
+- name: 'gcr.io/cloud-builders/docker'
+  dir: 'app'
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/myimage', '.']
+  id:   'build'
+images: ['gcr.io/$PROJECT_ID/myimage']
+```
--- a/docker-compose/cloudbuild.yaml
+++ b/docker-compose/cloudbuild.yaml
@@ -1,0 +1,10 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/docker-compose', '.']
+- name: 'gcr.io/$PROJECT_ID/docker-compose'
+  args: ['version']
+
+images: ['gcr.io/$PROJECT_ID/docker-compose']


ImJasonH:
First of all, thanks for this contribution! 

Unfortunately, our process doesn't currently allow us to accept PRs directly, so if we accept this change we'll have to apply it internally and mirror it out to GitHub, but we'd obviously reference this PR and attribute it to you when we do so.

I have two overall comments:

1.) I'd like to understand the use case for this a bit better. It sounds like the goal is to be able to have a series of build steps like:

```
steps:
# Build an image
- name: gcr.io/cloud-builders/docker
  args: ['build', '--tag=gcr.io/$PROJECT_ID/myimage', '.']
# Start the image in a local devel cluster
- name: gcr.io/cloud-builders/docker-compose
  args: ['up', 'docker-compose.yml']
# Run tests against local devel cluster
- name: ubuntu
  args: ['./run-tests.sh']
# Bring down the local devel cluster
- name: gcr.io/cloud-builders/docker-compose
  args: ['down']

# Push the locally tested image
images: ['gcr.io/$PROJECT_ID/myimage']
```

If that's the intention, that seems really useful. I personally don't know anything about `docker-compose`, but that example seems compelling if it's that easy.

Could you add an example under `/docker-compose/examples/something-here` that includes a sample image, a `docker-compose.yml` and some basic test script? It wouldn't have to do much, if it just started a server and tested that the server responded, that would be enough. Having a running example makes it easier to get started using it, and makes sure future releases don't break basic functionality.

2.) Would it be possible for the `docker-compose/Dockerfile` to be `FROM gcr.io/cloud-builders/docker`, so they stay in sync and share the same docker version? This reduces versioning complexity, since we only have one pair of versions at a time.

It looks like it's not possible to install `docker-compose` using `apt-get`, which is unfortunate, but ultimately not something either of us can do anything about. :-/

Anyway, this looks like a great change, and if we can address these points I think we could accept this change. If not, if this is something you and others find useful you can always build, run and publish this builder yourself, outside of `cloud-builders`.

Thanks again for the PR!

skippy:
hi @ImJasonH

thanks for the detailed response.  In order:

> Unfortunately, our process doesn't currently allow us to accept PRs directly

no problem!  

> 1.) I'd like to understand the use case for this a bit better...

the first few sentences on the docker-compose [documentation](https://docs.docker.com/compose/overview/) site provides a good overview.  it is a great way to run, test, and build an app when multiple containers are needed.  Think a dev env where your bit of code may need a mysql and redis container to run in dev mode and for tests.

Everything in docker-compose you can run in docker directly; it just provides a wrapper and some sugar to make it easier to manage (link containers; bring them up/down together, etc).

But your example is more complicated than need be.  The one I gave in the readme is 'good enough' ; like docker, running `docker-compose run` will retrieve and/or build the needed images. `docker-compose run app rake test` (for example) will build then run the app container defined in the `docker-compose.yml`, and if there are dependencies such as a mysql and redis container, those will be launched first.

I could provide a `docker-compose.yml` example file, such as: 
```yaml
version: '2'
services:
  db:
    image: postgres
  web:
    build: .
    command: mix phoenix.server
    volumes:
      - .:/app
    ports:
      - "4000:4000"
    depends_on:
      - db
```

but I think it is much better to just link to the docker-compose docs rather than put an example format that will be out of date sometime this year (version 3 of the DSL is forthcoming).  Anyone who uses docker-compose will know it's formats.


> 2.) Would it be possible for the docker-compose/Dockerfile to be FROM gcr.io/cloud-builders/docker

according to [build-steps docs](https://cloud.google.com/container-builder/docs/api/build-steps) the docker socket is linked to the call; all docker-compose does is send docker commands down the socket to the underlying process; so using `gcr.io/cloud-builders/docker` as the base image will include a bunch of binaries that will never be used.  Does that make sense?

skippy:
@ImJasonH sorry to hijack my own PR, but I'm not sure where to send feedback on [Container Builder](https://cloud.google.com/container-builder/docs/overview), but I have some feature ideas as well as a bug I may have found.  I'll enumerate them here, but if there is a better channel, or you want to followup offline, shoot me a note.  thanks!

* bug: using `waitFor` triggered a schema validation error; I tried all sorts of variations, used a yaml linter to make sure it wasn't a tab vs whitespace, etc, and no luck.  I could have made a bone-headed mistake, but I tried enough variations that I'm not sure that I did.

* feature request: the github integration is nice (!), but having it hook into the PR check functionality would be even nicer!!  This is a common feature of most build pipelines, and it provides a nice visual indicator on github that the PR was able to build (and pass tests) successfully

* post image processing: I was hoping to, after the image was processed, run a kubectl cmd to push the new image to my kubernetes cluster.  There are any number of post-processing steps that would be fantastic.  Now, there are many ways to skin this cat, as they say (create a process on the kubernetes cluster to watch the pub/sub and trigger it myself; or use a cron-job on the kubernetes cluster to see if their is a new image, and if so, trigger a rolling restart, etc.), BUT having post-processing steps in general is very nice and pretty normal for a builder.  THOUGH, the pub/sub is a good initial step

ImJasonH:
> I think it is much better to just link to the docker-compose docs rather than put an > example format that will be out of date sometime this year (version 3 of the DSL is > forthcoming). Anyone who uses docker-compose will know it's formats.

So the reason I want a working example is twofold: one, so that users can have a base to work from to write their own builds, and two, so that we can prove the builder works at all. If our documented example doesn't work, users might be put off and never use it, and we can't make sure it works unless we test it. A runnable example is much better than an example that "should" work.

If there's reason to believe that the usage of docker-compose will change, that might indicate that we should provide versioned builders, like we do with the docker builder, which has both `:1.9.1` and `:1.12.6`. We don't want users who had previously been happy to use the `docker-compose` builder to suddenly have it break in a future release.

> all docker-compose does is send docker commands down the socket to the underlying process

In that case it sounds like we don't need docker installed to run docker-compose? If so, maybe we don't need to base this on the `docker` builder. Am I understanding that correctly?

ImJasonH:
Re: your other comments:

> waitFor

We discovered a bug around inconsistent support for `waitFor` between `gcloud` and build triggers this past week, and have a fix in the works that should go out soon. Until then, using `wait_for` should work; please let us know if you find that not to be the case. In time, both variations of the field should be supported equally.

> GitHub PRs

This is something we'd definitely like to support, and we have some plans to enable it in time, but nothing immediately to announce.

> post image processing

We don't currently have a strong story for this in place. You could imaging having a build step that pushes, then a step that deploys, etc., but that's pretty messy, and it's not something we recommend.We had been focusing on the build(/test) story, and now that we have that, deploy is definitely something we're interested in supporting well, for k8s and others. Until then, relying on Pub/Sub is probably the best way forward.

skippy:
@ImJasonH thanks for the updates!  Is there a general-purpose open bug tracker?  I think I may have found another issue with cloud-builder, specifically with variable substitution (e.g. $BRANCH_NAME).  I'm testing it in detail now to see if I can find the boundaries of the inconsistent behavior.

skelterjohn:
https://github.com/GoogleCloudPlatform/cloud-builders/issues :)

skippy:
haha... ok, I wasn't sure that was for the whole project or just this repo!  I'll open up an issue once I do some more debugging.  thanks @skelterjohn

vorozhko:
Hi @skippy,
there is a video from @kelseyhightower who describe how to deploy to kubernetes using google cloud builder. See it here https://www.youtube.com/watch?v=XL9CQobFB8I

Best

skippy:
thanks @vorozhko!  I'll check that out.  I ended up doing a simple shell script that runs as part of a k8s cronJob; I'll see what @kelseyhightower .  

I'm new to the google cloud universe, and I have to say, gcloud is a *fantastic* cli; it makes stuff like this super simple

skippy:
hey @ImJasonH 

I'll throw in a sample for testing; we do have a test check in there, but a more complete test wouldn't be a bad thing.

I don't see it as a likely scenario (at all actually) that someone would use a `docker-compose.yml` just for this project; the flow tends to be use it in dev, and then perhaps ci.  but I'm arguing in the weeds; pardon :)  

Sample file with a test is forthcoming!

ahmetb:
fwiw I prefer this to be an alpine-based image rather than ubuntu to pull faster.

avinassh:
[from](https://github.com/GoogleCloudPlatform/cloud-builders/pull/10#issuecomment-285975145)

> You could imaging having a build step that pushes, then a step that deploys, etc., but that's pretty messy, and it's not something we recommend.

Why it is not recommended?

bendory:
It is not recommended because we envision Container Builder as a build tool, not a deployment orchestration tool. So while it is possible -- given the flexibility of build steps -- to use Container Builder for orchestration and workflow, we recommend instead using an actual deployment orchestration workflow tool such as http://spinnaker.io; such tools are designed for that.

ahmetb:
@avinassh  To be fair, you can use https://hub.docker.com/r/docker/compose/ as a step to make deployments, you shouldn't need an official cloud-builders image from Google. 

> It is not recommended because we envision Container Builder as a build tool, not a deployment orchestration tool.

@bendory There's already demand for `kubectl apply` (#2), so it's not very hard to imagine people will use it the way they like and there’s not really a good way to prevent people from doing that. 

`<personal-opinion>`
Spinnaker is really complicated (and overkill) for pretty much everybody and sometimes a `kubectl apply` is all you need to roll out a new version of your application. Kubernetes has built-in rolling updates and saves you from using Spinnaker for basic use cases.
`</personal-opinion>`

bendory:
Indeed, we have seen requests to make `kubectl` [#22](https://github.com/GoogleCloudPlatform/cloud-builders/issues/22), `gcloud app deploy` [#36](https://github.com/GoogleCloudPlatform/cloud-builders/issues/36), and other deployment mechanisms available in Container Builder. We are considering a new feature to enable such build steps, but we still don't plan to recommend Container Builder as an orchestration or workflow tool at this time. Watch [#36](https://github.com/GoogleCloudPlatform/cloud-builders/issues/36) for updates on the enabling feature.

DocBradfordSoftware:
@bendory, I just don't know what to say about your vision of Cloud Builder, it almost just leaves me speechless. Why would google make a big deal about a product that is just a glorified "gcloud push" is beyond me.  What good is building images and storing them in Google Storage, if you can't automatically use it?

bendory:
You most certainly can automatically use the pushed image -- indeed, in any way you want if you listen for GCR notifications and take action in response to them. We're simply focused here on a Container Builder. We've made it composable and very flexible so that you can use it in a whole variety of ways, including those under discussion here.

Philmod:
Here is an example to run integration tests on GCB using docker-compose: https://github.com/Philmod/gcb-docker-compose

Philmod:
Hey @skippy, sorry I closed this PR too fast. 
Could you update the version of the tool and add an example (like for other builders)?

Philmod:
Hi @skippy, any chance you could update this PR? Otherwise I will have to recreate it.

Philmod:
We created another repository for community builders, the docker-compose is there: https://github.com/GoogleCloudPlatform/cloud-builders-community

PR #15, opened 2017-03-15
Update go/.dockerignore to exclude the README

ahmetb:


bendory:
Unfortunately, our process doesn't currently allow us to accept PRs directly, so I'm applying this change internally and will mirror it out to GitHub. I've referenced this PR and attributed it to you.

PR #18, opened 2017-03-16
Add example for Bazel builder

ImJasonH:
This simple example, based on
https://bazel.googlesource.com/bazel/+/master/tools/build_defs/docker/README.md,
produces a Docker image that simply runs "echo foo" based on debian.

### DIFF
--- a/bazel/examples/debian.BUILD
+++ b/bazel/examples/debian.BUILD
@@ -1,0 +1,15 @@
+load("@bazel_tools//tools/build_defs/docker:docker.bzl", "docker_build")
+
+# Extract .xz files
+genrule(
+    name = "wheezy_tar",
+    srcs = ["tianon-docker-brew-debian-e9bafb1/wheezy/rootfs.tar.xz"],
+    outs = ["wheezy_tar.tar"],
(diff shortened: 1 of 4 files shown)


bazel/examples/debian.BUILD
```diff
@@ -0,0 +1,15 @@
+load("@bazel_tools//tools/build_defs/docker:docker.bzl", "docker_build")
+
+# Extract .xz files
+genrule(
```
skelterjohn: could we add a custom rule, included with our bazel builder, that let you specify, eg
docker_tar(
  name = "wheezy_tar",
  src = "https://github.com/tianon/docker-brew-debian/tarball/e9bafb113f432c48c7e86c616424cb4b2f2c7a51",
  sha256 = "d31d8215e336a011d9a5594634595c6cb390ef90fa72a51c0043c5e7c3fd5ba1",
)

or something?

ImJasonH: Possibly, I don't know. This is just taken directly from the Bazel documentation. If we/users find some improvement to the Docker workflow in Bazel, I think we'd have more impact by moving that into Bazel's Docker support directly.

PR #20, opened 2017-03-19
Add support for docker-engine 1.13.1

ben-davis:
- Updates latest to 1.13.1
- Allows use of --cache-from

Tested by manually building the docker builder to a private project and running a new dependent build.

### DIFF
--- a/docker/Dockerfile-1.13.1
+++ b/docker/Dockerfile-1.13.1
@@ -1,0 +1,19 @@
+FROM ubuntu:trusty
+
+# Based on instructions from:
+# https://docs.docker.com/engine/installation/linux/ubuntu/
+RUN \
+   apt-get -y update && \
+   apt-get -y install apt-transport-https ca-certificates curl \
+       # These are necessary for add-apt-respository
+       software-properties-common python-software-properties && \
+   curl -fsSL https://yum.dockerproject.org/gpg | sudo apt-key add - && \
+   apt-key fingerprint 58118E89F3A912897C070ADBF76221572C52609D && \
+   add-apt-repository \
+       "deb https://apt.dockerproject.org/repo/ \
+       ubuntu-$(lsb_release -cs) \
+       main" && \
+   apt-get -y update && \
+   apt-get -y install docker-engine=1.13.1-0~ubuntu-trusty
+
+ENTRYPOINT ["/usr/bin/docker"]
--- a/docker/README.md
+++ b/docker/README.md
@@ -6,7 +6,7 @@ Arguments passed to this builder will be passed to `docker` directly, allowing
 callers to run [any Docker
 command](https://docs.docker.com/engine/reference/commandline/).
 
-By default, the version of Docker that is used by this builder is `1.12.6`.
+By default, the version of Docker that is used by this builder is `1.13.1`.
 
 ## GCR Credentials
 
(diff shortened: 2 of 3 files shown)


ben-davis:
I signed it!

Philmod:
Hey Ben,

We are currently still using the version 1.12.6 in the backend. 

In order to evaluate the upgrade, could you tell me what you are looking for in the 1.13.1 version?

Cheers,
Philmod

ben-davis:
Hey Phil - I was hoping to take advantage of `--cache-from` from 1.13 to attempt to bring down my build times. Currently it takes > 10 mins because of Python dependencies on `gensim`, `numpy`, and `scipy`. As I understand it, Container Builder spins up a new VM on each build and therefore doesn't allow for layer caching. I was hoping that manually pulling the latest image and using it with `--cache-from` would fix this, although currently it hasn't affected anything and the image still builds from scratch. Any advice on how to fix this would be very appreciated.

Obviously this was a very quick PR to put together to test, so if it's not of any use I'll happily close it.

Philmod:
As you understood correctly, every build is started on a fresh VM.

I'm not familiar with the `--cache-from` yet, but as I understand you would have to pull an image to use the cache from, which might not make the build faster.
Could you tell me how it would help in this case?
Have you benchmarked it locally?

ben-davis:
Yeah you're right, you pull a specific version of an image and pass it to `--cache-from`. Pulling the image takes about 20 seconds, but the build takes > 10 minutes because of dependencies that take a while to build, so in this case it would save a lot of time.

Locally it still takes a while if it's the first build, but any subsequent builds are much fast due to local Docker caching.

skelterjohn:
Upgrading the docker client won't do anything for --cache-from without us
upgrading the daemon as well. We're necessarily a little conservative with
our docker upgrades, so please be patient :)

ben-davis:
@skelterjohn okay that makes sense, I'll close this then. Is the upgrade on your roadmap? It's quite difficult for us to use Container Builder as a part of CI process (as much as we want to) unless we can find a way around the massive build time.

skelterjohn:
I'll contact you directly to talk about ways to speed up your build.

ben-davis:
Great, thanks @skelterjohn

skelterjohn:
I find that github no longer offers an easy way to direct-message someone. So, I might as well talk here and then everyone else can see, too.

Would it be possible to create an intermediate base image? That is, instead of

    # build to runtime-product
    FROM base
    RUN some stuff you want cached
    RUN build your source change

have

    # build to intermediate-base
    FROM base
    RUN some stuff you want cached

    # build to runtime-product
    FROM intermediate-base
    RUN build your source change

This approach also gives you more direct control over when your cache is refreshed.

ben-davis:
Yeah that could work. But where would the logic live that decided when the base image needed to rebuilt? Surely it's the responsibility of a build system to decide when different elements  that comprise that build need to rebuilt? I suppose it depends on how much Container Builder should be considered a CI tool. But if I have to maintain a separate CI tool to handle the orchestration of Container Builder, well then I may as well use that CI tool to do the actual building of the images.

bendory:
Please note that as of last week, Container Builder is running on Docker engine 17.05.

ben-davis:
I can confirm that with 17.05, using `--cache-from` now allows cloud build to use cached layers.

Issue #1, opened 2017-01-13
go: document libc and ca-certificates.crt dependencies?


evanj:
When applying the examples for the `go` builder to my project, I ran into two problems:

1. By default, the binary depends on libc (some version of musl it seems?). This can be worked around by adding `env: ["CGO_ENABLED=0"]`.
2. If you make HTTPS connections, it needs `/etc/ssl/certs/ca-certificates.crt` for the trusted root certificates.

I think this is mostly a documentation oversight. I think it could be "solved" in any of the following ways:

* Document these dependencies/limitations in `go/README.md`?
* Recommend that people use the `golang-project` builder and not `go`. I think `golang-project` sets an alpine base image in the Dockerfile by default, which would avoid this.
* Document the alpine base image that is used by the builder, so we could add it as the correct base instead of `scratch`?
* Ship a minimal base image that includes the correct `libc` version and `ca-certificates.crt` in the correct location, so we could use that instead of alpine (probably overkill; this makes tiny binaries, but adds maintenance burden).
* Add a working example that connects to https://www.google.com, since I think that runs into both these dependencies (for the resolver and to verify the HTTPS certificate).

skelterjohn:
Thanks for the feedback, evanj.

I've already got something in the works for CGO_ENABLED=0, I'll try to roll that out soon.

For the cert issue, I'm not sure how I can address that in the go builder, which only creates a binary which relies on its container to have certs.

I like your suggestion about the example, I'll take care of it.

evanj:
Fair point: Re: certificates: It is up to the user to configure their image correctly; I just wish it was somehow more obvious that the binary alone is not sufficient, that said: there are lots of guides about including it. Maybe an easy fix is to have the examples reference a more complete base image instead of scratch, so that if someone like me just copies an example and tries to use it, its more likely to work correctly? A user who wants to cut down their image size is then on their own (or can reference the various Internet guides about it).

libc is more difficult: with the current container builder, its looking for some version of `musl`, and it isn't clear how to find the right one. Disabling cgo works, but is also not ideal: I have things that use sqlite, and it might be nice to use the cloud builder with them.

Anyway, thanks, feel free to close this as you see fit.

skelterjohn:
Check out the changes just pushed and let me know what you think. They're live in production now.

I'm also considering having a go:alpine and go:debian as separate builders - then we'd avoid the CGO_ENABLED=0 requirement and you could have your sqlite fun.

[...]

PR #24, opened 2017-03-21
Add Gulpjs as a cloud build option

amaxwell01:
* Added Gulpjs - http://gulpjs.com/
* Added gulp version command for a test ```gulp --version```
* Added gulp help command for a test ```gulp --help```
* Created a simple example gulp build file to test ```gulp build```

### DIFF
--- a/gulpjs/Dockerfile
+++ b/gulpjs/Dockerfile
@@ -1,0 +1,6 @@
+# Use the base cloud builder NPM
+FROM gcr.io/cloud-builders/npm
+
+RUN npm install -g gulp-cli
+
+ENTRYPOINT ["gulp"]
--- a/gulpjs/cloudbuild.yaml
+++ b/gulpjs/cloudbuild.yaml
@@ -1,0 +1,31 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+# Step 0
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/gulpjs', '.']
+
+# Step 1
+# Print gulp version.
+- name: 'gcr.io/$PROJECT_ID/gulpjs'
+  args: ['--version']
+
+# Step 2
+# Gulp --help
+- name: 'gcr.io/$PROJECT_ID/gulpjs'
+  args: ['--help']
+
+# Step 3
+# Install gulp into the example build project
+- name: 'gcr.io/cloud-builders/npm'
+  args: ['install']
+  dir: 'examples/build'
+
+# Step 4
+# Gulp build
+- name: 'gcr.io/$PROJECT_ID/gulpjs'
+  args: ['build']
+  dir: 'examples/build'
+
+images: ['gcr.io/$PROJECT_ID/gulpjs']
--- a/gulpjs/examples/build/package.json
+++ b/gulpjs/examples/build/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "Gulpjs build",
+  "version": "1.0.0",
+  "description": "Gulpjs build test",
+  "main": "gulpfile.js",
+  "license": "Apache License 2.0",
+  "dependencies": {
+    "gulp": "^3.9.1"
+  }
+}
--- a/gulpjs/.gitignore
+++ b/gulpjs/.gitignore
@@ -1,0 +1,2 @@
+# Don't upload the node modules folder
+node_modules
--- a/gulpjs/README.md
+++ b/gulpjs/README.md
@@ -1,0 +1,9 @@
+# Tool builder: `gcr.io/cloud-builders/gulpjs`
+
+This Container Builder build step runs the `gulp` tool.
+
+## Building this builder
+
+To build this builder, run the following command in this directory.
+
+    $ gcloud container builds submit . --config=cloudbuild.yaml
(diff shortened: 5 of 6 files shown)


gulpjs/Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Use the base App Engine Docker image, based on debian jessie.
+FROM debian:8
```
Philmod: Why not using the npm builder as the base?

gulpjs/cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
...
+# Install gulp into the example build project
+- name: 'gcr.io/cloud-builders/npm'
+  args: ['install']
+  dir: 'gulpjs/examples/build'
+
+# Step 4
+# Gulp build
+- name: 'gcr.io/$PROJECT_ID/gulpjs'
+  args: ['build']
+  dir: 'gulpjs/examples/build'
```
Philmod: That would be nice to have an example that produces a static file, and in a next step pushes it to a CDN or to GCS.
I believe that's a realistic use case of gulpjs?

gulpjs/examples/build/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "build",
+  "version": "1.0.0",
+  "description": "",
+  "main": "gulpfile.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
Philmod: Either you remove this script, or you add a npm test as a additional build step (but then don't `exit 1`).

gulpjs/examples/build/package.json
```diff
@@ -0,0 +1,14 @@
...
+  },
+  "author": "",
+  "license": "ISC",
```
Philmod: Please use the Apache License. https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/LICENSE

gulpjs/cloudbuild.yaml
```diff
@@ -0,0 +1,32 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+# Step 0
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/gulpjs', '.']
+  dir: 'gulpjs'
```
skelterjohn: this file is meant to be executed from this directory, not the top of the repository (omit "gulpjs" from the `dir` fields).

JustinBeckwith:
I think @GoogleCloudPlatform/node-team should take a look!

gulpjs/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# Install updates and dependencies
+RUN apt-get update -y && apt-get install --no-install-recommends -y -q curl python build-essential git ca-certificates libkrb5-dev imagemagick && \
+    apt-get clean && rm /var/lib/apt/lists/*_*
+
+# Install the latest LTS release of nodejs
+RUN mkdir /nodejs && curl https://nodejs.org/dist/v6.9.1/node-v6.9.1-linux-x64.tar.gz | tar xvzf - -C /nodejs --strip-components=1
+ENV PATH $PATH:/nodejs/bin
+RUN npm install -g gulp-cli
```
JustinBeckwith: @jmdobry @ofrobots 

Typically people will install gulp as a dev dependency in their package.json.  If you install dev dependencies, it also symlinks the entrypoint so you can just invoke `gulp` in an npm script.  This is important for producing hermetic builds, and also so you can pin to a specific version of gulp that's not head.  We should never `npm install -g` a dev dependency like this for a user.  

On other words - I don't think this is the right way to make this builder at all.  A proper builder for node.js can be built with a single multi-purpose pipeline. 

Proposed Steps
- Run `npm install` (notice that I didn't add --production).  
- Run some user defined build step, possibly by convention:  `npm run build`
- Build the container
- Entrypoint is set to `npm start`

If we're feeling super fancy we could do a two container process where we trim the npm dependencies down to only the --prod ones in the actual runtime container. 

@ofrobots - this really intersects with the builder work y'all are already doing.  We should sync.

amaxwell-gpsw: Agreed that maybe pinning it to head is not the best, we could switch to specific versions. 

Right now there are multiple ways in the community to trigger gulp builds within a project:
1. "make" routing to the node_module dependency command
2. npm script
3. running the script via a node file or command line

The goal that I was trying to make was to abstract the different use cases for developers so that they can simplify their build steps. For my teams we usually use #1 but that requires both a dev dependency install and a prod install(two steps). Abstracting the gulp from the build avoids the use of installing dev dependencies if you wanted to do something simple using gulp.

amaxwell01:
Addressed most of the feedback except for:

```
That would be nice to have an example that produces a static file, and in a next step pushes it to a CDN or to GCS.
I believe that's a realistic use case of gulpjs?
```

Which is something that I can explore tonight.

gulpjs/Dockerfile
amaxwell-gpsw: I'd love to hop on a call or chat if you guys are interested.

Philmod: We are going to discuss it internally first. I'll come back to you.
If you could provide a real example, that would be very helpful.

Philmod:
Hey Andrew,

I wrote an example on how to use Gulpjs with GCB: https://github.com/Philmod/google-container-gulp

Feel free to reach out for any question.

amaxwell-gpsw:
The code looks good @Philmod nicely done. Better example to use the package specific version of Gulp.

PR #29, opened 2017-03-28
Add a `.bazelrc` to configure the `--output_base`

mattmoor:
This change adds a default `--output_base` to the bazel builder to make it store its outputs under `/workspace`.  This is because what lives under `/workspace/bazel-bin/` is actually just a set of symlinks to the actual outputs, so downstream build steps get to see the symlinks, but not the things to which those symlinks refer.

By moving the output based under `/workspace`, the symlinks will correctly resolve.

### DIFF
--- a/bazel/Dockerfile
+++ b/bazel/Dockerfile
@@ -35,4 +35,8 @@ RUN \
     # Unpack bazel for future use.
     bazel version
 
+# Store the Bazel outputs under /workspace so that the symlinks under bazel-bin (et al) are accessible
+# to downstream build steps.
+RUN echo 'startup --output_base=/workspace' > ~/.bazelrc
+
 ENTRYPOINT ["bazel"]


Philmod:
This PR broke the bazel builder.

ImJasonH:
Specifically:

```
Step #2: Already have image: gcr.io/argo-integration-testing/bazel
Starting Step #2
Step #2: INFO: Reading 'startup' options from /root/.bazelrc: --output_base=/workspace
Step #2: ..................
Step #2: Blaze should not be called from a Blaze output directory. The pertinent workspace directory is: '/workspace/examples'
Finished Step #2
```

mattmoor:
@damienmg Guidance?  Has this recently changed?

I swear I tested this, perhaps we should add a `.cache` or something to the path?

I'm doing the equivalent thing [here](https://github.com/google/containerregistry/blob/master/pusher.yaml#L7)...

Rollback for now?

damienmg:
Setting the output base somewhere under the workspace directory is dangerous so we made it an error, can't you move the project to some place underneath workspace?

mattmoor:
I don't understand what you mean by "underneath workspace"?

Argo has `/workspace` volume mounted in and is the root of the build (where the Git repo is cloned).  In order to access the resulting build artifacts after Bazel is invoked, the symlinks must reference stuff within that volume, thus putting the outputs under `/workspace`.

damienmg:
Then you should probably move the root of the build to workspace/root so
output_base is not underneath the bazel workspace.

skelterjohn:
I don't think it's reasonable for us to require Cloud Builder + Bazel users to rearrange their repositories in what appears to be a very artificial way.

damienmg:
No user should have to change their workspace, why? Why are you forced to git clone in the root of the workspace?

skelterjohn:
I didn't mean to imply that bazel should change, here. Only that I don't like the /workspace/root solution. We'll find another one.

The source for your build is put in /workspace, and /workspace is the directory that is persisted from step to step. Perhaps we could also persist /tmp?

mattmoor:
Perhaps in addition to `/workspace` each container may export volumes, which are mounted into subsequent containers with `--volumes-from <prior container>` ?

skelterjohn:
That sounds like a complicated DX :)

PR #45, opened 2017-04-20
Add a .bazelrc to configure the --output_base

mattmoor:
This is a second attempt at #29, which broke builds where the Bazel workspace was a subdirectory of `/workspace`.  According to the Bazel folks, things should be fine if we give Bazel a dedicated subdirectory as `--output-base`.

### DIFF
--- a/bazel/Dockerfile
+++ b/bazel/Dockerfile
@@ -35,4 +35,8 @@ RUN \
     # Unpack bazel for future use.
     bazel version
 
+# Store the Bazel outputs under /workspace so that the symlinks under bazel-bin (et al) are accessible
+# to downstream build steps.
+RUN echo 'startup --output_base=/workspace/.bazel' > ~/.bazelrc
+ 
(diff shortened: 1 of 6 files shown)


mattmoor:
@ImJasonH @skelterjohn Let's try this again :)

ImJasonH:
Is there an example build that we can include in bazel's cloudbuild.yaml to demonstrate that this is working correctly?

mattmoor:
I `gsutil` a build artifact to GCS, but the way this is generically written that is non-trivial.

We could probably build a trivial app and then run a stock `ubuntu` container in which we run: `/workspace/bazel-bin/foo`...

ImJasonH:
Yeah, that would be great. I just want to make sure some future change doesn't break this, since I'm probably the most likely to do that. :)

mattmoor:
@ImJasonH done

ImJasonH:
Perfect, thanks!

PR #55, opened 2017-05-19
Use the newly built docker:17.05 to tag :latest

skelterjohn:
Otherwise, the old version requires -f to overwrite a tag. And on a future version, -f would not be allowed. So, we pin this step to a 17.05 to get deterministic behavior.


bendory:
LGTM

PR #60, opened 2017-05-26
Build a matrix of java/mvn builders.

skelterjohn:
Build a matrix of java/mvn builders, crossing maven 3.5.0/3.3.9 with java 7/8.

Use build args to parameterize the base image. Copypaste in cloudbuild.yaml. Possibly generate it with a script in the future.

### DIFF
--- a/java/mvn/Dockerfile
+++ b/java/mvn/Dockerfile
@@ -1,4 +1,5 @@
-FROM openjdk:8-jdk-alpine
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
 
 RUN apk add --update --no-cache curl tar bash
 
--- a/java/mvn/Dockerfile
+++ b/java/mvn/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE
+ARG BASE_IMAGE=openjdk:8-jdk-alpine
 FROM ${BASE_IMAGE}
 
 RUN apk add --update --no-cache curl tar bash
--- a/java/mvn/cloudbuild.yaml
+++ b/java/mvn/cloudbuild.yaml
@@ -45,7 +45,28 @@ steps:
   args: ['tag', 'gcr.io/$PROJECT_ID/java/mvn:3.5.0-jdk-8', 'gcr.io/$PROJECT_ID/java/mvn']
 
 # run examples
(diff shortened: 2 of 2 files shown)


bendory:
LGTM

java/mvn/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM openjdk:8-jdk-alpine
+ARG BASE_IMAGE
```
skelterjohn: This needs a default value.

java/mvn/cloudbuild.yaml
```diff
@@ -2,8 +2,45 @@
...
+  - '--build-arg=BASE_IMAGE=openjdk:7-jdk-alpine'
+  - '--build-arg=MAVEN_VERSION=3.5.0'
+  - '--build-arg=SHA=beb91419245395bd69a4a6edad5ca3ec1a8b64e41457672dc687c173a495f034'
+  - '--tag=gcr.io/$PROJECT_ID/java/mvn:3.5.0-jdk-7'
+  - '.'
+  wait_for: ['-']
+
+# Tag a default builder version.
 - name: 'gcr.io/cloud-builders/docker'
   args: ['tag', 'gcr.io/$PROJECT_ID/java/mvn:3.5.0-jdk-8', 'gcr.io/$PROJECT_ID/java/mvn']
```
aslo: Does it make sense to run the examples with more of these new images? Or at least with the :latest image?

java/mvn/cloudbuild.yaml
```diff
@@ -2,8 +2,45 @@
...
```
aslo: This step only needs to `waitFor` the step that builds the `3.5.0-jdk-8` image, right?

skelterjohn: Technically that is true, but it takes no time so I didn't bother making it run concurrently.

java/mvn/cloudbuild.yaml
skelterjohn: done

aslo: Thoughts on parallelizing running each of the examples?

skelterjohn: Since they share /workspace I was worried about them clobbering each other in flight. Is that not a concern?

aslo: Ah, right. Good point.

PR #63, opened 2017-06-04
Simplify docker build using --build_arg.

bendory:
- build all versions with a single Dockerfile using --build-arg to pass DOCKER_VERSION.
- let Docker cache the common layers leading up to the actual Docker install.
- run steps concurrently.


docker/cloudbuild.yaml
```diff
@@ -4,45 +4,82 @@
 steps:
 # Build all supported versions.
 - name: 'gcr.io/cloud-builders/docker'
```
ImJasonH: If we want these to build concurrently, these should each have "waitFor: ['-']", otherwise 1.12.6 won't start until 1.9.1 is done.

skelterjohn:
LGTM

docker/cloudbuild.yaml
bendory: Yeah, I considered making these concurrent, then didn't bother. Can add in a new PR if you want...

PR #69, opened 2017-06-13
Use docker-credential-gcr for docker authentication.

skelterjohn:


skelterjohn:
Note that this won't go into effect until we stop volume-mounting the credential file into build steps, since it depends on the contents of that file remaining fixed as they are in the image.

bendory:
Note: @skelterjohn and I discussed off-line; there are a number of unresolved issues with this PR that need to be resolved prior to merge.

ImJasonH:
Is this PR still active?

skelterjohn:
nope!

Issue #22, opened 2017-03-20
How do I run kubectl after image push?


DocBradfordSoftware:
It would seem that a CI/CD pipeline would require the ability to deploy an image into the gke cluster after it was built/tested/pushed.

Without this, it seems like Builder is incomplete.

skelterjohn:
At the moment, you will have trouble using `kubectl` from within the container builder service, since the credentials used do not have the proper scopes. We will address this issue soon.

For pushing before running another step, you can always run `docker push` as its own step (either instead of or in addition to listing the image in the `images` field).

    steps:
    - name: 'gcr.io/cloud-builders/docker'
      args: ['build', '-t', 'gcr.io/$PROJECT_ID/foo', '.']
    - name: 'gcr.io/cloud-builders/docker'
      args: ['push', 'gcr.io/$PROJECT_ID/foo']
    - name: 'gcr.io/cloud-builders/kubectl'
      args: ['run', 'foo-pod', '--image=gcr.io/$PROJECT_ID/foo']
    images: ['gcr.io/$PROJECT_ID/foo']

(for an upcoming `kubectl` step that is waiting on the scope issue)

DocBradfordSoftware:
Yes, that would work.  Any idea if the scopes have to be added to the cluster and will it be able to be done after the fact.
I Built the gke/jenkins integration and I had to create a new cluster with the proper scope.

skelterjohn:
The scopes I'm referring to, which are used to create credentials for your builder service account (https://cloud.google.com/container-builder/docs/permissions), don't apply to any resources outside of the container builder service. So, scopes you use with your cluster are unrelated.

What's missing at this point is the ability to use the GKE cluster API to list your clusters and get the auth secrets necessary for talking to the cluster master. Once those secrets are acquired, you can use `kubectl` to talk to the cluster master directly (and schedule pods, etc).

bussyjd:
I am very interested in this capability.
When can we expect to have the scope issue solved? Thanks a lot.

Philmod:
The other solution is to create a Google Cloud Function that listens to the `cloud-builds` PubSub feed, and update the appropriate deployment of your Kubernetes cluster. 

In this [example](https://gist.github.com/Philmod/290fdabf386646dbbc0b57b00952b944), I have two repositories that map to two different deployments in a same namespace.

DocBradfordSoftware:
Well, that's pretty cool, @Philmod, that should work in the interim.

[...]

PR #70, opened 2017-06-15
Add a kubectl build step.

skelterjohn:
Fixes https://github.com/GoogleCloudPlatform/cloud-builders/issues/22

### DIFF
--- a/gcloud/Dockerfile
+++ b/gcloud/Dockerfile
@@ -17,8 +17,9 @@ RUN apt-get -y update && \
         --bash-completion=false \
         --disable-installation-options && \
 
-    # Install alpha and beta components
-    /builder/google-cloud-sdk/bin/gcloud -q components install alpha beta && \
+    # Install additional components
+    /builder/google-cloud-sdk/bin/gcloud -q components install \
+        alpha beta kubectl && \
     /builder/google-cloud-sdk/bin/gcloud -q components update && \
 
     # install crcmod: https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod
--- a/kubectl/README.md
+++ b/kubectl/README.md
@@ -3,6 +3,8 @@
 This Container Builder build step runs
 [`kubectl`](https://kubernetes.io/docs/user-guide/kubectl-overview/).
 
+## Using this builder with Google Container Engine
+
 To use this builder, your [builder service account]
 (https://cloud.google.com/container-builder/docs/how-to/service-account-permissions)
 will need IAM permissions sufficient for the operations you want to perform. For
@@ -15,14 +17,14 @@ For most use, kubectl will need to be configured to point to a specific GKE
 cluster. You can configure the cluster by setting environment variables.
 
     CLOUDSDK_COMPUTE_ZONE=<your cluster's zone>
-    CLOUDSDK_CONTAINER_CLUSTER=<your cluster name>
+    CLOUDSDK_CONTAINER_CLUSTER=<your cluster's name>
 
 Setting the environment variables above will cause this step's entrypoint to
 first run a command to fetch cluster credentials as follows.
 
     gcloud container clusters get-credentials --zone "$CLOUDSDK_COMPUTE_ZONE" "$CLOUDSDK_CONTAINER_CLUSTER"`
 
-Then, `kubectl` will have 
+Then, `kubectl` will have the configuration needed to talk to your GKE cluster.
 
 ## Building this builder
 
--- a/kubectl/README.md
+++ b/kubectl/README.md
@@ -1,0 +1,31 @@
+# Tool builder: `gcr.io/cloud-builders/kubectl`
+
+This Container Builder build step runs
+[`kubectl`](https://kubernetes.io/docs/user-guide/kubectl-overview/).
+
+To use this builder, your [builder service account]
+(https://cloud.google.com/container-builder/docs/how-to/service-account-permissions)
+will need IAM permissions sufficient for the operations you want to perform. For
+typical read-only usage, the "Container Engine Viewer" role is sufficient. To
+deploy container images on a GKE cluster, the "Container Engine Developer" role
+is sufficient. Check the [GKE IAM page]
+(https://cloud.google.com/container-engine/docs/iam-integration) for details. 
+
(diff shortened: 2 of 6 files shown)


kubectl/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+    CLOUDSDK_COMPUTE_ZONE=<your cluster's zone>
+    CLOUDSDK_CONTAINER_CLUSTER=<your cluster name>
+
+Setting the environment variables above will cause this step's entrypoint to
+first run a command to fetch cluster credentials as follows.
+
+    gcloud container clusters get-credentials --zone "$CLOUDSDK_COMPUTE_ZONE" "$CLOUDSDK_CONTAINER_CLUSTER"`
+
+Then, `kubectl` will have
```
bendory: this looks incomplete?

kubectl/README.md
```diff
@@ -0,0 +1,31 @@
...
+typical read-only usage, the "Container Engine Viewer" role is sufficient. To
+deploy container images on a GKE cluster, the "Container Engine Developer" role
+is sufficient. Check the [GKE IAM page]
+(https://cloud.google.com/container-engine/docs/iam-integration) for details. 
+
+For most use, kubectl will need to be configured to point to a specific GKE
+cluster. You can configure the cluster by setting environment variables.
+
+    CLOUDSDK_COMPUTE_ZONE=<your cluster's zone>
+    CLOUDSDK_CONTAINER_CLUSTER=<your cluster name>
```
bendory: for consistency: "your cluster's name"

kubectl/README.md
```diff
@@ -0,0 +1,31 @@
...
+will need IAM permissions sufficient for the operations you want to perform. For
+typical read-only usage, the "Container Engine Viewer" role is sufficient. To
+deploy container images on a GKE cluster, the "Container Engine Developer" role
+is sufficient. Check the [GKE IAM page]
+(https://cloud.google.com/container-engine/docs/iam-integration) for details. 
+
+For most use, kubectl will need to be configured to point to a specific GKE
+cluster. You can configure the cluster by setting environment variables.
+
+    CLOUDSDK_COMPUTE_ZONE=<your cluster's zone>
```
ImJasonH: Does kubectl also take args to set these?

kubectl/kubectl.bash
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	[[ -z "$cluster" ]] && var_usage
+	[[ -z "$zone" ]] && var_usage
+	err=$(mktemp)
+	gcloud container clusters get-credentials --zone "$zone" "$cluster" 2> "$err"
+	if [[ "$?" != 0 ]]; then
+		cat "$err"
+		exit 1
+	fi
```
bendory: This block seems like the equivalent of
`gcloud container clusters get-credentials --zone "$zone" "$cluster" || exit`
except that it swallows stderr on success. I think just running the command is simpler (and potentially provides diagnostic error details even on success?)

kubectl/cloudbuild.yaml
```diff
@@ -0,0 +1,6 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/kubectl', '.']
+
```
ImJasonH: Is there anything we can do to smoke test the image? "kubectl version" even?

gcloud/Dockerfile
```diff
@@ -17,8 +17,9 @@ RUN apt-get -y update && \
         --bash-completion=false \
         --disable-installation-options && \
 
-    # Install alpha and beta components
-    /builder/google-cloud-sdk/bin/gcloud -q components install alpha beta && \
+    # Install additional components
+    /builder/google-cloud-sdk/bin/gcloud -q components install \
+        alpha beta kubectl && \
```
bendory: Help me understand why we install kubectl in the gcloud builder rather than just adding it to the kubectl builder?

kubectl/kubectl.bash
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# If there is no current context, get one.
+if [[ $(kubectl config current-context 2> /dev/null) == "" ]]; then
+	cluster=$(gcloud config get-value container/cluster 2> /dev/null)
+	zone=$(gcloud config get-value compute/zone 2> /dev/null)
+	function var_usage() {
+		cat <<EOF
+No cluster is set. To set the cluster (and the zone where it is found), set the environment variables
```
ImJasonH: Does kubectl itself fail with a helpful error message if these values aren't found? It might be better to use the common error path, in case the requirements change in a future release. kubectl's error message will also presumably have more helpful search results, since all anybody ever does is google error messages.

gcloud/Dockerfile
skelterjohn: Fewer layers, and we're caching the data anyway.

kubectl/README.md
skelterjohn: No, this are configurations for `gcloud container clusters get-credentials`, which does take them both as args. `kubectl` itself has no concept of GKE.

kubectl/README.md
skelterjohn: done

kubectl/README.md
skelterjohn: oops, thanks

kubectl/cloudbuild.yaml
skelterjohn: done

kubectl/kubectl.bash
skelterjohn: kubectl has no concept of GKE and cannot help you configure it to point to a GKE cluster.

kubectl/kubectl.bash
skelterjohn: To me, it seemed spammy, especially without seeing the command run. I could put a message about what it's doing first?

bendory: I prefer the simpler code with an "echo get-credentials" message first, but YMMV.

kubectl/kubectl.bash
ImJasonH: Oh I get it, these values aren't actually passed to kubectl in any way, they're only used to call `get-credentials` and put creds where `kubectl` can find them. Got it, thanks.

PR #87, opened 2017-07-12
Update java/mvn builder's environment variables

aslo:
This change updates a few of the maven builder's env variables:
* `MAVEN_HOME` is changed to `M2_HOME`. Even though these images all use Maven 3, `M2_HOME` is actually the [preferred environment variable name](https://stackoverflow.com/questions/17136324/what-is-the-difference-between-m2-home-and-maven-home) for Maven versions 2 and later.
*  `MAVEN_CONFIG` is removed. This was of questionable usefulness in the first place, and it causes issues when using this image with [popular maven wrapper scripts](https://github.com/takari/maven-wrapper/blob/master/mvnw#L225). We'll address wrapper support more thoroughly in the future, but I still think it's reasonable justification for this change.


aslo:
cc @skelterjohn, @meltsufin

Issue #88, opened 2017-07-12
Move default dependency resolution to internal google maven mirror


adamresson:
I noticed the discussion in #50 by @aslo and @skelterjohn 

Does it make sense to point to the internal google cloud maven mirror described here:
https://www.infoq.com/news/2015/11/maven-central-at-google

I can work up a PR is someone thinks it's worthwhile. While it's nice to have a few 'predefined' dependencies, pointing to what should be a 'faster' maven server seems to make sense in my mind.

skelterjohn:
It's in our plans. Perhaps your issue will encourage @aslo to bump some other things to get it done ;)

And PRs are always welcome - cc @aslo on PRs for java builders.

aslo:
@adamresson Yep! This is something that I'm currently looking into :)

Right now there's some performance testing that's underway to ensure that pulling dependencies from the GCP maven mirror is indeed as speedy as we expect. Once we confirm, we'll update this builder image.

PR #89, opened 2017-07-12
Point to internal Google Maven mirror

adamresson:
I took a shot at this @aslo based on our discussion from #88 

I actually tried this locally and it immediately failed to build the docker image
```
[INFO] Downloading: https://maven-central.storage.googleapis.com/org/apache/maven/plugins/maven-dependency-plugin/3.0.1/maven-dependency-plugin-3.0.1.pom
[WARNING] The POM for org.apache.maven.plugins:maven-dependency-plugin:jar:3.0.1 is missing, no dependency information available
[INFO] Downloading: https://maven-central.storage.googleapis.com/org/apache/maven/plugins/maven-dependency-plugin/3.0.1/maven-dependency-plugin-3.0.1.jar
```

Is there something special about that repo what doesn't allow any 'external' connectivity?

### DIFF
--- a/java/mvn/Dockerfile
+++ b/java/mvn/Dockerfile
@@ -17,6 +17,9 @@ RUN mkdir -p /usr/share/maven /usr/share/maven/ref \
 
 ENV M2_HOME /usr/share/maven
 
+# Point maven central to google internal mirror
+ADD settings.xml $USER_HOME_DIR/.m2/settings.xml
+
 # transitively resolve all dependencies
 ADD deps.txt /builder/deps.txt
 ADD resolve-deps.sh /builder/resolve-deps.sh
--- a/java/mvn/settings.xml
+++ b/java/mvn/settings.xml
@@ -1,0 +1,10 @@
+<settings>
+  <mirrors>
+    <mirror>
+      <id>google-maven-central</id>
+      <name>Google Maven Central</name>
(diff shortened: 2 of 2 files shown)


aslo:
The correct mirror URL is slightly different - if you update your settings.xml so that it looks like what I have below, it should work. However, there are some network performance issues that I'd like to resolve before we merge this change. Otherwise, this may not actually make builds much faster at all!

```
<settings>
  <mirrors>
    <mirror>
      <id>google-maven-central</id>
      <name>Google Maven Central</name>
      <url>https://maven-central.storage-download.googleapis.com/repos/central/data</url>
      <mirrorOf>central</mirrorOf>
    </mirror>
  </mirrors>
</settings>
```

danielcompton:
@aslo have these issues been resolved? Is it safe to use these mirrors?

aslo:
Hi @danielcompton, I haven't worked on this for a while, but I've been told that the mirrors should now be at least as fast as Maven central from within GCP. 

@lesv may be able to provide some more detail.

lesv:
The last time I tested we were running about the same speed or slightly faster than maven-central.

lesv:
There was an issue with freshness, that appears to have been resolved.  I still need a bit of instrumentation to make sure it doesn't happen again.

PR #91, opened 2017-07-13
Implement cloud-builders/java/gradle build step

aslo:
Implements a [gradle](https://gradle.org/) builder, across a matrix of jdk and gradle releases. This PR also includes an example gradle application which is built as a sort of "sanity check" in the cloud build execution for these images.

### DIFF
--- a/java/gradle/Dockerfile
+++ b/java/gradle/Dockerfile
@@ -1,0 +1,25 @@
+ARG BASE_IMAGE=openjdk:8-jdk-alpine
+FROM ${BASE_IMAGE}
+
+RUN apk add --update --no-cache curl tar bash
+
+ARG GRADLE_VERSION=4.0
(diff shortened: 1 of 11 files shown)


aslo:
cc @meltsufin

java/gradle/Dockerfile
```diff
@@ -0,0 +1,25 @@
+ARG BASE_IMAGE=openjdk:8-jdk-alpine
+FROM ${BASE_IMAGE}
+
+RUN apk add --update --no-cache curl tar bash
```
skelterjohn: Do we need curl, tar, and bash when people use the build step? Otherwise you could merge this RUN into the one that downloads and installs gradle, and delete it in the same RUN.

aslo: Done.

PR #97, opened 2017-07-18
Java images use debian-based images

### LINKED ISSUE #94: Install libc6-compat in the alpine-based Java build steps
This library is required by gRPC. We'll save users a lot of time by just installing it ourselves.

### LINKED ISSUE #96: Java builders based on alpine
Nearly all of our builders are based on `ubuntu` by default, unless some other basis is necessary, or unless we provide multiple builders based on `ubuntu` and some other distro (e.g., `golang-project:ubuntu` and `golang-project:alpine`).

Alpine generally produces smaller images, but it also means users have to use `ash` instead of `bash`, and in general know more about differences between Alpine and Ubuntu/Debian. Smaller builder images are nice, but are ultimately less meaningful since worker VMs pre-cache our official builders before receiving builds.

Unless there's some reason `java/mvn` and `java/gradle` must be based on `openjdk:8-jre-alpine` I think we should consider basing them on `openjdk:8-jre` for consistency, and to continue the precedent set by existing official builders. This image is based on `buildpack-deps:stretch-curl` which is based on `debian:stretch`.

If they must be based on `openjdk:8-jre-alpine` we should document somewhere why we made that decision to help future builder developers understand it.

aslo:
fixes #96 
fixes #94

Debian-based images make more sense for our use case. This PR updates all Java images to stop using alpine.

### DIFF
--- a/java/gradle/Dockerfile
+++ b/java/gradle/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=openjdk:8-jdk-alpine
+ARG BASE_IMAGE=openjdk:8-jdk
 FROM ${BASE_IMAGE}
 
 ARG GRADLE_VERSION=4.0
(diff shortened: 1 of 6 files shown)


ImJasonH:
Nit: These images are based on `debian:stretch` and not `ubuntu`

aslo:
:+1: I updated the descriptions. I still think these base images make the most sense though.

PR #98, opened 2017-07-18
Add a go builder example that uses GCE metadata

skelterjohn:
Add a go builder example that uses GCE metadata to make a GCE request, without setting the GCE_METADATA_HOST environment variable.

### DIFF
--- a/go/examples/gcp_test/Dockerfile
+++ b/go/examples/gcp_test/Dockerfile
@@ -1,0 +1,1 @@
+FROM alpine
--- a/go/examples/gcp_test/README.md
+++ b/go/examples/gcp_test/README.md
@@ -1,0 +1,6 @@
+# gco_test
+
+This example build creates binary that uses officially supported Go libraries
+to talk to the build's GCE metadata service and make GCE API calls.
+
+The API call this build performs lists GCE instances in all zones.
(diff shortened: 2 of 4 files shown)


go/examples/gcp_test/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	projectID, err := metadata.ProjectID()
+	if err != nil {
+		log.Fatalf("Could not get Project ID: %v", err)
+	}
+
+	client := oauth2.NewClient(ctx, google.ComputeTokenSource(""))
```
ImJasonH: I believe the compute.New call by default sets itself up to use Application Default Credentials, which means it'll hit metadata in the absence of any other option.

go/examples/gcp_test/README.md
```diff
@@ -0,0 +1,6 @@
+# gco_test
```
ImJasonH: gcp_test

go/examples/gcp_test/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM alpine
```
ImJasonH: What's the point of this empty Dockerfile?

go/examples/gcp_test/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+	gce, err := v1compute.New(client)
+	if err != nil {
+		log.Fatalf("Could not create GCE client: %v", err)
+	}
+
+	resp, err := gce.Instances.AggregatedList(projectID).Do()
```
ImJasonH: Could you list builds instead? Or just do anything that doesn't require an additional permission that the builder service account doesn't have by default?

PR #103, opened 2017-08-03
Add support for current node version 8.4.0 in npm and bump LTS to 6.11.2

tlbdk:
This updates the npm build containers to a similar setup as the docker containers so it builds multiple versions, supporting both LTS and current of node

### DIFF
--- a/npm/Dockerfile
+++ b/npm/Dockerfile
@@ -6,7 +6,8 @@ RUN apt-get update -y && apt-get install --no-install-recommends -y -q curl pyth
     apt-get clean && rm /var/lib/apt/lists/*_*
 
 # Install the latest LTS release of nodejs
-RUN mkdir /nodejs && curl https://nodejs.org/dist/v6.9.1/node-v6.9.1-linux-x64.tar.gz | tar xvzf - -C /nodejs --strip-components=1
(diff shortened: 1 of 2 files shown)


tlbdk:
I signed it!

Philmod:
I like that you keep the `latest` version the one that exists today, to avoid any breakage.

Adding @JustinBeckwith to the thread who is responsible of the nodejs language.

tlbdk:
I updated the build so it creates two images, one keeping the old version with the latest tag and a new one with 8.2.1 getting the current tag, so everything should work as before if you don't update your tag, also I tried to do it the same ways as the docker build was done.

Philmod:
Thought: should we use the npm version, not the nodejs one?

tlbdk:
I was thinking the same first, but the issue is that the node versions often means more than the npm version, fx our tests run fine on an old npm version as it has change very little but require newer node features. This is the more common use case. Also new node version will always ship with a new default npm version so the idea here would be to follow whatever node is shipped with.

tlbdk:
I bump the versions to newest versions for both current and LTS

Philmod:
The new images have been built: `docker pull gcr.io/cloud-builders/npm:node-8.4.0`

PR #105, opened 2017-08-04
Implement java/openjdk base container with docker

aslo:
This PR creates a common openjdk base image which contains an installation of docker. Eventually, we plan to use this image as the base image for the other java tool builders (`java/mvn` and `java/gradle`).


aslo:
@skelterjohn A few notes/questions:

- I didn't set an entrypoint in the java/openjdk image, nor did I add example projects, as I was thinking this image would mainly just serve as a base for other tool builders. Does that sound right to you? Or do you think this base image should double as a `javac` tool builder or something like that?
- I'm thinking that switching to use this as a base image for `java/mvn` and `java/gradle` should happen in a separate PR, after this image is released. Does that sound right to you?

aslo:
I'm not clear on what the use case for a "java" tool builder would be. If there's demand for one we can easily add it later. Sound good?

skelterjohn:
Yes, that sounds fine to me.

The use would be if someone wanted to run a jar file inside a build step. They could use an available JRE image and make sure to use the right entrypoint/args, so perhaps not necessary.

PR #106, opened 2017-08-07
Adding `dotnet` builder.

ivannaranjo:
This PR adds a `dotnet` builder that allows the use to invoke .NET Core's `dotnet` command line tool in their `cloudbuild.yaml` scripts. This `dotnet` builder contains support for the 1.0.0 and 1.0.4 SDKs.

The PR also contains an example WebApi app, with the corresponding `cloudbuild.yaml` and `Dockerfile` that demonstrate how to build an app's image using this builder.


dotnet/README.md
```diff
@@ -0,0 +1,16 @@
+# Tool builder: gcr.io/cloud-builders/dotnet
+This directory contains the description for the `dotnet` Container Build step. This step will run the .NET Core `dotnet` command line. This builder has two SDKs installed:
```
skelterjohn: "This builder has two SDKs installed:"

This is fine, but there is precedence for having the version of the tool or SDK in the image tag. So, people would use the 'gcr.io/cloud-builders/dotnet:1.0.3' builder.

If this is a thing commonly configured through CLI flag or project config, it makes sense to have it like you do here.

dotnet/README.md
```diff
@@ -0,0 +1,16 @@
...
++ 1.0.3 with SDK Preview 2 build 3156, to support .NET Core 1.0.5 with project.json solutions.
++ 1.0.4, which supports .csproj projects for .NET Core 1.0.5 and 1.1.2.
+
+Any arguments passed to this builder step will be passed directly to the `dotnet` tool. Note that when using the `dotnet restore` command you should always specify the `--packages` option to a path within the `/workspace`, this way the restored packages will be available to the next invocations of `dotnet`. See the [example](examples/TestApp/cloudbuild.yaml) for how use `dotnet resotre` correctly.
```
skelterjohn: > `/workspace`

Can we avoid mentioning /workspace? Does it make sense to give a relative path as the flag value?

skelterjohn: `dotnet resotre` -> restore

dotnet/README.md
skelterjohn: I see later that you advice just that. SGTM.

dotnet/cloudbuild.yaml
```diff
@@ -0,0 +1,14 @@
+# Cloud builder for the dotnet builder.
+steps:
+- name: gcr.io/cloud-builders/docker
+  args: [ 'build', '-t', 'gcr.io/$PROJECT_ID/dotnet', '.' ]
+
+# Print versions information
```
skelterjohn: Is this to ensure that it's processing the config files properly? Can you make a step that checks that output?

```
  name: 'gcr.io/$PROJECT_ID/dotnet'
  entrypoint: 'bash'
  dir: 'configs/config-1.0.0'
  args: ['-c', '[[ $(dotnet --version) == "expected output" ]]']
```

dotnet/cloudbuild.yaml
```diff
@@ -0,0 +1,14 @@
+# Cloud builder for the dotnet builder.
+steps:
+- name: gcr.io/cloud-builders/docker
```
skelterjohn: please quote all string fields in yaml. I know it works either way, but I'd like all string fields to be consistent.

dotnet/README.md
ivannaranjo: It does and in the cloudbuild.yaml file it uses a relative path. In this document I mention /workspace to explain why we need to do that. Perhaps I should reword the doc.

dotnet/cloudbuild.yaml
ivannaranjo: Clever :), I'll add that.

dotnet/cloudbuild.yaml
ivannaranjo: Done.

ivannaranjo:
Addressed the comments, added a toplevel `csharp` directory to contain the `dotnet` builder. PTAL.

dotnet/cloudbuild.yaml
ivannaranjo: Done.

dotnet/cloudbuild.yaml
ivannaranjo: Done.

dotnet/README.md
ivannaranjo: Done.

PR #108, opened 2017-08-11
javac builder: keep apt-transport-https package

aslo:
Updates the Dockerfile of the `javac` build step to keep the `apt-transport-https` package. 

This is necessary because if `apt-transport-https` is removed, the package manager is left in a bad state for any downstream images, since it contains an https repository, but no means of interacting with it. This means that most `apt-get` calls will fail, making it difficult to reinstall the required package and fix the problem.

Also considered alternative: removing the docker ppa. However, this makes it difficult for downstream images to update to newer versions of the `docker-ce` apt package.

### DIFF
--- a/java/javac/Dockerfile
+++ b/java/javac/Dockerfile
@@ -18,7 +18,7 @@ RUN \
    apt-get -y install docker-ce=${DOCKER_VERSION} && \
 
    # Clean up build packages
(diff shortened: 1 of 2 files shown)


aslo:
cc @meltsufin PTAL

meltsufin:
LGTM

PR #109, opened 2017-08-14
Yarn lives in nodejs subdir; clean up example.

bendory:


Philmod:
Could you test the new example image to make sure installing yarn also installs nodejs and other requirements?

Philmod:
@bendory Were you able to run this example? 
I had to fix the npm example too: https://github.com/GoogleCloudPlatform/cloud-builders/pull/110

ImJasonH:
This isn't necessary after #156

PR #115, opened 2017-08-17
Use cloud-builders/yarn builder in yarn example

ImJasonH:
Using the $PROJECT_ID/yarn builder might fail if the project doesn't have that builder yet.


nodejs/yarn/examples/hello_world/cloudbuild.yaml
```diff
@@ -1,8 +1,9 @@
 steps:
-- name: 'gcr.io/$PROJECT_ID/yarn'
+- name: 'gcr.io/cloud-builders/nodejs/yarn'
```
Philmod: Don't we want to keep the subdirectly `nodejs` before `yarn`?

ImJasonH: I added a second commit, it's gcr.io/cloud-builders/nodejs/yarn

PR #137, opened 2017-09-27
add gradle 4.2

wyaeld:


wyaeld:
oops, wasn't support to open against the official parent, sorry

PR #149, opened 2017-10-26
add docker to gcloud builder

erain:
* gcloud's docker sub-commands requires docker installed in the system,
  therefore the current gcloud builder doesn't work for docker related
  functions

https://cloud.google.com/container-registry/docs/pushing-and-pulling

### DIFF
--- a/gcloud/Dockerfile
+++ b/gcloud/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:trusty
+FROM gcr.io/cloud-builders/docker:latest
 
 RUN apt-get -y update && \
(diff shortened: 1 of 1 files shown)


ImJasonH:
I don't believe this change is necessary. I think `gcloud docker` is only provided as a wrapper to pass Google credentials to Docker commands, which the `gcr.io/cloud-builders/docker` builder already takes care of, without the need for the `gcloud` wrapper.

If there's some reason you've found you need to invoke `gcloud docker` instead of just `docker`, this might indicate a bug, and please feel free to open an issue.

PR #153, opened 2017-11-10
Move java builders to top-level directories

ImJasonH:
This removes java/* and builds new images at `gcr.io/$PROJECT_ID/{mvn,gradle,javac}` in addition to their `java/` counterparts.

We'll recommend users use the top-level builder images, and eventually when the java/ variants aren't used anymore, we can stop building them and delete them.


nof20:
LGTM.  Thanks!

PR #154, opened 2017-11-17
Add gcc to the go builder.

bendory:


bendory:
Note: Needed this to support the Rocker community builder -- https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/6

bendory:
Further note: adding gcc to the go builder enabled a simple `go get` to build packages like `rocker` that include gcc compilation.

PR #155, opened 2017-12-04
Add deprecation notices for golang-project

ImJasonH:


golang-project/README.md
```diff
@@ -1,3 +1,7 @@
+# !!! `golang-project` is deprecated !!!
+
+**Use `gcr.io/cloud-builders/go` instead**
+
```
bendory: Do we need to be so friendly as to give instructions for how to "convert" from golang-project to the vanilla go builder?

ImJasonH: Sure, I'll add those instructions in a future PR.

PR #157, opened 2017-12-04
Move csharp/dotnet builder to top-level dotnet directory

ImJasonH:


dotnet/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
+# Cloud builder for the dotnet builder.
+#
+# TODO(franklinn): Stop tagging csharp/dotnet images once usage has dropped off.
```
bendory: on github, he's nof20

PR #159, opened 2017-12-06
Add instructions to migrate off of golang-project

ImJasonH:


ImJasonH:
Added golang-project/examples/hello_world_migrated which is the hello_world example, migrated off of golang-project.

PR #165, opened 2017-12-11
use openjdk image maintained by java runtimes team

### LINKED ISSUE #164: Use gcr.io/google-appengine/openjdk instead of dockerhub openjdk?
Are you guys open to using the openjdk image that the cloud java runtimes team maintains instead of the dockerhub version? (https://github.com/GoogleCloudPlatform/openjdk-runtime)

We use your mvn image, and noticed that if your base openjdk image was switched to the one that Google maintains then we could also monitor the image for security/vulnerabilities (https://github.com/GoogleCloudPlatform/runtime-builder-java/issues/92)

ChengyuanZhao:
fixes #164

### DIFF
--- a/javac/Dockerfile
+++ b/javac/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=gcr.io/google-appengine/openjdk:8
+ARG BASE_IMAGE=launcher.gcr.io/google/openjdk8
 FROM ${BASE_IMAGE}
 
 ARG DOCKER_VERSION=17.06.0~ce-0~debian
--- a/javac/Dockerfile
+++ b/javac/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=openjdk:8-jdk
+ARG BASE_IMAGE=gcr.io/google-appengine/openjdk:8
 FROM ${BASE_IMAGE}
 
 ARG DOCKER_VERSION=17.06.0~ce-0~debian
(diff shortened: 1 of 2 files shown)


ChengyuanZhao:
@ImJasonH @meltsufin PTAL

meltsufin:
Does anyone still need JDK 7 builders? JDK 7 has been EOL since 2015!

javac/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=openjdk:8-jdk
+ARG BASE_IMAGE=gcr.io/google-appengine/openjdk:8
```
meltsufin: I would suggest using the mirror at `launcher.gcr.io/google/openjdk8` because otherwise it might raise questions of why "appengine"?

@dlorenc FYI

ImJasonH: +1 to using the launcher tag

ImJasonH:
It looks like there's very little usage of the JDK7 builders (for javac, mvn or gradle) so sunsetting them might be the right solution. This is ultimately the decision of that Java runtimes team though @meltsufin

meltsufin:
We do not provide a vanilla OpenJDK 7 image. If you want to use the one from DockerHub, it's fine. Otherwise, my recommendation would be to just remove Java 7 builders.

ImJasonH:
Filed #167 to track sunsetting JDK7 builders

meltsufin:
LGTM

ChengyuanZhao:
FYI I can't merge this in myself because of lack of access.

PR #180, opened 2017-12-19
Upgrade node and npm version to newest current and LTS

tlbdk:


Philmod:
@JustinBeckwith Should we still build the older versions?

tlbdk:
We could add all supported versions and tag them with exact and major version:

Fx these tags:
4.8.7 and 4.x
6.12.2 and 6.x
8.9.3, 8.x and latest
9.3.0 9.x and current

https://github.com/nodejs/Release

ofpiyush:
Can we merge this in? Waiting to use 8.9.0. Would love to cloud builder images as they mean shorter build times. :)

chainlink:
+1 Would also love this

ImJasonH:
I think it makes sense to keep building old versions for a bit. That way if someone is broken by 9.3 they can at least specify 8.4 to get the old "current" builder.

@tlbdk if you want to make that change in this PR that would be best, otherwise I'll merge this and send another PR shortly after to add the old latest/current versions.

ofpiyush:
@ImJasonH the current versioned images will continue to be present and usable

Are you worried about updates to the underlying OS images etc?

ImJasonH:
Yeah we'll keep building the old versions ~daily to pick up any underlying
changes, until we decide to stop and untag the image entirely.

tlbdk:
@ImJasonH I won't have time to look at this the next couple of days so if you have time it would be great

ImJasonH:
See #192 which adds the new versions but doesn't update `:latest` and `:current` just yet.

PR #186, opened 2018-01-05
Update Bazel example

ImJasonH:
- Use `go_image` rule to build an image directly from Go source
- Remove newly unnecessary `docker_pull`
- Move example source up from `subdir` and remove unnecessary `go_prefix`
- Use `container_image` rule instead of (equivalent) `docker_build` rule

/cc @mattmoor


bazel/examples/WORKSPACE
```diff
@@ -17,18 +17,15 @@ git_repository(
...
-    name = "base_image",
-    registry = "index.docker.io",
-    repository = "library/debian",
-    digest = "sha256:d3892ebd85bb21b161e9d170beefd6c444d9bd58aba3998e1f67aa842fbedb9d", # wheezy
+load(
+    "@io_bazel_rules_docker//container:container.bzl",
+    container_repositories = "repositories",
 )
+
+container_repositories()
```
mattmoor: technically this is unnecessary with the `_go_image_repos()` (they all call this common dep).

PR #189, opened 2018-01-08
Stop building gcr.io/$PROJECT_ID/docker:1.12.6

ImJasonH:


ImJasonH:
Just cleaning up an unused version.

Issue #188, opened 2018-01-06
Docker 17.09+ is required to use --chown flag on ADD/COPY operations in Dockerfile


ryan-mf:
We're trying to speed up our builds, and one of the ways to do that is to pull in the lastest image tag and use it as a cache source during the docker build.

However, our Dockerfile uses pip install with a requirements.txt, and we're trying to use the newer --chown flag when our Dockerfile copies in the requirements.txt to solve a CI issue where differing permissions inside CI vs inside GCB results in cache invalidation.

Given that the pip install step is the vast majority of our build time (and completely redundant on every build given our requirements.txt), we're hoping to get GCB up to Docker 17.09 ASAP so we can start using the --chown flag to ignore the source permissions as a means to keep the cache valid.

Philmod:
Hey Ryan,
We are planning to upgrade Docker version in Q1.

bendory:
Rejecting in favor of #194

codrienne:
Docker has been upgraded to 17.12.0-ce.

PR #194, opened 2018-01-16
Add docker 17.12.0-ce

sebglon:
Create new Dockerfile for docker 17.12

#188

### DIFF
--- a/docker/cloudbuild.yaml
+++ b/docker/cloudbuild.yaml
@@ -52,8 +52,8 @@ steps:
 # Tag the latest version as :latest. We use gcr.io/cloud-builders/docker here
 # and not gcr.io/$PROJECT_ID/docker because the latter may not yet exist.
 - name: 'gcr.io/cloud-builders/docker'
-  args: ['tag', 'gcr.io/$PROJECT_ID/docker:17.12.0', 'gcr.io/$PROJECT_ID/docker']
-  wait_for: ['17.12.0']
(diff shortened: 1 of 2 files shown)


bendory:
Hi @sebglon, thanks for the PR. We have begun looking at the upgrade to 17.12, but it is a bit more complex than this PR in that we need to first get the Docker server upgrade into place.

Is there a particular feature driving your request for the upgrade?

sebglon:
yes using --chown  has describe on #188

docker/cloudbuild.yaml
```diff
@@ -32,13 +40,20 @@ steps:
...
+  wait_for: ['17.12.0']
+
 # Tests for future supported versions of docker builder go here.
 
 # Tag the latest version as :latest. We use gcr.io/cloud-builders/docker here
 # and not gcr.io/$PROJECT_ID/docker because the latter may not yet exist.
 - name: 'gcr.io/cloud-builders/docker'
-  args: ['tag', 'gcr.io/$PROJECT_ID/docker:17.06.1', 'gcr.io/$PROJECT_ID/docker']
-  wait_for: ['17.06.1']
+  args: ['tag', 'gcr.io/$PROJECT_ID/docker:17.12.0', 'gcr.io/$PROJECT_ID/docker']
```
ImJasonH: We can't tag `17.12` as `:latest` until the server upgrade is complete, and until we've tested that the new client version doesn't break anything.

`17.12` will be broken until the server upgrade, but simply building+pushing `17.12` is fine, since nobody is forced to use it (until it's tagged `:latest`).

codrienne:
Docker has been updated to 17.12.0.

PR #197, opened 2018-01-19
Add bzip2 so PhantomJS can be installed

jak:
The npm package for phantomjs downloads a bzip2 file. In the old version of this image, this worked because `build-essentials` depends on `dpkg-dev` which made `bzip2` available.

### DIFF
--- a/yarn/Dockerfile
+++ b/yarn/Dockerfile
@@ -3,7 +3,7 @@ FROM launcher.gcr.io/google/debian8
 
 # https://yarnpkg.com/en/docs/install#linux-tab
 RUN apt-get update && \
-    apt-get install -y curl apt-transport-https git && \
+    apt-get install -y curl apt-transport-https git bzip2 && \
 
     # Install newer Node version.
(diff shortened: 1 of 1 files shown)


jak:
I signed it!

ImJasonH:
Thanks! Sorry this yarn upgrade has been so rough :-/

haf:
Is this patch gone again? What package does now provide bzip2?

bendory:
@haf Did this just disappear? Or did you happen to just notice?

I ask because it looks to me like bzip2 was (presumably accidentally) deleted here: https://github.com/GoogleCloudPlatform/cloud-builders/pull/241/files

bendory:
Hmm, investigating further, it looks like bzip2 is actually installed:

> $ docker run --rm --entrypoint=/bin/bash gcr.io/cloud-builders/yarn:latest -- which bzip2
> /bin/bzip2
>

jak:
As referenced in my original message, `build-essential` depends on `bzip2` (indirectly!). Now that `build-essential` is back in, it shouldn't be an issue.

PR #204, opened 2018-01-26
Update yarn tool/container build to push tag

### LINKED ISSUE #203: Missing image tags
I'd like our builds to be repeatable.

We use the `yarn` tool/container. We had a build breaking due to https://github.com/GoogleCloudPlatform/cloud-builders/commit/2d9d102aeb9d286663d337b1394ce31e1ff04bc3 because of an upgrade of the node version (it's been locked down in our repo and npm crashes with that error, so it's really doing what it's intended to do).

Could you please tag your tools so we can use a specific version, rather than having to depend on `:latest`? Like, tagging `:v1.3.2` in this case.

haf:
Fixes #203

### DIFF
--- a/yarn/cloudbuild.yaml
+++ b/yarn/cloudbuild.yaml
@@ -32,4 +32,5 @@ steps:
 
 images:
 - 'gcr.io/$PROJECT_ID/yarn'
+- 'gcr.io/$PROJECT_ID/yarn:1.3.2'
 - 'gcr.io/$PROJECT_ID/nodejs/yarn'
--- a/yarn/cloudbuild.yaml
+++ b/yarn/cloudbuild.yaml
@@ -8,6 +8,8 @@ steps:
   args:
   - 'build'
   - '--tag=gcr.io/$PROJECT_ID/yarn'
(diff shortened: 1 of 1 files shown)


Philmod:
You would need to also push the new tag. Look at the `npm` builder: https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/npm/cloudbuild.yaml

Philmod:
You would need to also push the new tag. Look at the `npm` builder: https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/npm/cloudbuild.yaml

haf:
@Philmod Done

Philmod:
Great! 

Could you tell me why you picked this version? Is this the LTS version?

haf:
I picked the version you have on your master since 7 days back. Yarn has been stable for me since ~0.14 when I started using it (contrary to most other software I've used ;)), so I don't care that much about LTS. Upgrades between yarn versions mostly affect format of the lockfile as far as I've seen.

Philmod:
Fair enough :)

PR #207, opened 2018-02-01
Updating dotnet builders 2 0 5 and 1 1 6 and 1 0 9

ivannaranjo:
Updating to the latest SDK releases for .NET Core.


dotnet/Dockerfile
```diff
@@ -20,19 +20,19 @@ RUN apt-get update \
...
 # + 1.0.3 with SDK Preview 2 build 3156, to support .NET Core 1.0.5 with project.json solutions.
-# + 1.1.5, which supporst .csproj projects for .NET Core 1.0.8 and 1.1.5.
-# + 2.1.3, which supports .csproj projects for .NET Core 2.0.4.
+# + 1.1.7, which supports .csproj projects for .NET Core 1.0.9 and 1.1.6.
+# + 2.1.4, which supports .csproj projects for .NET Core 2.0.5.
 RUN mkdir -p /usr/share/dotnet && \
     curl -sL https://go.microsoft.com/fwlink/?LinkID=836302 | tar -xz -C /usr/share/dotnet && \
-    curl -sL https://download.microsoft.com/download/C/5/5/C55807F5-601C-49B1-B9BB-1BE03EB83E0A/dotnet-dev-debian-x64.1.1.5.tar.gz | tar -xz -C /usr/share/dotnet/ && \
-    curl -sL https://download.microsoft.com/download/2/9/3/293BC432-348C-4D1C-B628-5AC8AB7FA162/dotnet-sdk-2.1.3-linux-x64.tar.gz | tar -xz -C /usr/share/dotnet && \
+    curl -sL https://download.microsoft.com/download/4/E/6/4E64A465-F02E-43AD-9A86-A08A223A82C3/dotnet-dev-debian-x64.1.1.7.tar.gz | tar -xz -C /usr/share/dotnet/ && \
```
ImJasonH: Should we still build images that have 1.1.5 and 2.1.3 installed? Is there any concern that this upgrade might break users?

ImJasonH:
Also, is there any motivating reason for this upgrade? Some new feature supported in newer versions? Just trying to get an idea for the size/scope of this upgrade.

dotnet/Dockerfile
ivannaranjo: Yes, each SDK is backwards compatible with their previous major.minor build, so the `1.1.7` SDK is compatible with previous `1.1.x` SDKs and so on.

ivannaranjo:
This will keep the SDKs up to date with the latest releases of .NET. Nothing major simply keeping up with Microsoft's lifecycle.

ImJasonH:
Thanks Ivan. I expect this will be released on Monday.

PR #209, opened 2018-02-02
Add the GCR credential helper to the docker image.

mattmoor:
It is notable that this won't be used on GCB unless one of these is specified to the builder, since it overrides `HOME=/builder/home` today:
* `HOME=/root/`
* `DOCKER_CONFIG=/root/.docker`

### DIFF
--- a/docker/Dockerfile-17.05.0
+++ b/docker/Dockerfile-17.05.0
@@ -1,4 +1,4 @@
-FROM launcher.gcr.io/google/ubuntu16_04
+FROM docker-base:latest
 
 RUN apt-get -y update && \
     apt-get -y install \
--- a/docker/Dockerfile-17.06.1
+++ b/docker/Dockerfile-17.06.1
@@ -1,4 +1,4 @@
-FROM launcher.gcr.io/google/ubuntu16_04
+FROM docker-base:latest
 
 RUN apt-get -y update && \
     apt-get -y install \
--- a/docker/cloudbuild.yaml
+++ b/docker/cloudbuild.yaml
@@ -2,6 +2,13 @@
 # $ gcloud container builds submit . --config=cloudbuild.yaml
 
 steps:
+# Build a shared base with the credential-helper
+- name: 'gcr.io/cloud-builders/docker'
+  args:
+  - 'build'
+  - '--file=Dockerfile-base'
+  - '--tag=docker-base:latest'
+  - '.'
 # Build all supported versions.
 - name: 'gcr.io/cloud-builders/docker'
   args:
--- a/docker/Dockerfile-base
+++ b/docker/Dockerfile-base
@@ -1,0 +1,9 @@
+FROM launcher.gcr.io/google/ubuntu16_04
+
+ENV CRED_HELPER_VERSION=1.4.1
+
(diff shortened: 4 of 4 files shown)


mattmoor:
@ImJasonH FYI

docker/Dockerfile-17.05.0
```diff
@@ -15,4 +15,10 @@ RUN apt-get -y update && \
     apt-get -y update && \
     apt-get -y install docker-ce=17.05.0~ce-0~ubuntu-xenial
 
+ENV CRED_HELPER_VERSION=1.4.1
+
+RUN  curl -L https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${CRED_HELPER_VERSION}/docker-credential-gcr_linux_amd64-${CRED_HELPER_VERSION}.tar.gz  | tar xz -O docker-credential-gcr > /usr/local/bin/docker-credential-gcr && chmod +x /usr/local/bin/docker-credential-gcr
```
ImJasonH: Can we split this across lines for readability?

mattmoor: Done

docker/Dockerfile-17.06.1
```diff
@@ -1,4 +1,4 @@
-FROM launcher.gcr.io/google/ubuntu16_04
+FROM gcr.io/cloud-builders/docker-base:latest
```
ImJasonH: These should be `gcr.io/$PROJECT_ID/docker-base:latest` -- the intention is that any user can run these builds in their own project to get a `docker` builder in their own registry.

Does the name of the base image matter at all though? Could it just be `docker-base` since it's not pushed?

docker/cloudbuild.yaml
```diff
@@ -2,6 +2,14 @@
...
 
 steps:
+# Build a shared base with the credential-helper
+- name: 'gcr.io/cloud-builders/docker'
+  args:
+  - 'build'
+  - '--file=Dockerfile-base'
+  - '--tag=gcr.io/cloud-builders/docker-base:latest'
+  - '.'
+  id: 'base'
```
ImJasonH: Don't need an `id` if nothing `waitFor`s on it.

docker/Dockerfile-17.06.1
mattmoor: Done

ImJasonH:
Is there a way to make this work with the default overridden `HOME` that GCB applies today? After thinking about this, it seems a bit weird to add this path through the builder but hide it behind some configuration.

Today, GCB periodically rewrites `~/.docker/config.json` with fresh tokens, but it could be nice to have the default behavior read Google auth tokens from the metadata server and remove that special path.

mattmoor:
You could move to writing the cred helper info in `config.json`, but this would be breaking for anyone relying on these credentials without the cred helper (e.g. Bazel). :(

mattmoor:
Didn't realize this wasn't in.  What do you want to do here?

ImJasonH:
I was hoping there would be a way that every user of this builder could use the cred helper and metadata auth, instead of only having it enabled by setting magic env vars, which I don't think many users would actually do.

My (possibly over-cautious) policy with official builders, especially with one like `docker` which accounts for the vast majority of build steps we run, is to keep it as simple as possible, and relegate any improvements/complications to a non-official builder.

Having the builder always use metadata auth instead of periodically rewriting `config.json` certainly sounds like a vast simplification, but not if users have to know to set these envs to benefit. Or if the builder sets these envs, and has a different `$HOME` behavior than other steps.

I don't know the best way to proceed, basically.

PR #222, opened 2018-03-09
Add g++ and make

artjomzab:
Some packages like node-mongo require g++ and make


artjomzab:
I signed it!

Philmod:
This builder has `build-essential` already

PR #223, opened 2018-03-09
Upgrade docker to 17.12.0-ce

codrienne:
\o/


codrienne:
I am confused as to the long history of commits; there should only be 3.

PR #241, opened 2018-03-26
Update yarn cloud builder so node.js version is configurable

palmerj3:
Follow the same pattern as the npm cloud builder and allow for the Node.JS version to be configurable in the yarn cloud builder

PR #249, opened 2018-04-02
Allow node version to be configurable in yarn cloud builder

palmerj3:
Redo of https://github.com/GoogleCloudPlatform/cloud-builders/pull/241 which handles setting a default node version properly.

cc @codrienne

### DIFF
--- a/yarn/Dockerfile
+++ b/yarn/Dockerfile
@@ -1,16 +1,16 @@
 # Use the base App Engine Docker image, based on debian jessie.
 FROM launcher.gcr.io/google/debian8
 
-# https://yarnpkg.com/en/docs/install#linux-tab
-RUN apt-get update && \
-    apt-get install -y curl apt-transport-https git bzip2 build-essential && \
+# Install updates and dependencies
+RUN apt-get update -y && apt-get install --no-install-recommends -y -q curl python build-essential git ca-certificates libkrb5-dev imagemagick apt-transport-https && \
+    apt-get clean && rm /var/lib/apt/lists/*_*
 
-    # Install newer Node version.
-    curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
-    apt-get install -y nodejs && \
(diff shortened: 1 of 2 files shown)


yarn/Dockerfile
```diff
@@ -1,16 +1,16 @@
...
-    curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
-    apt-get install -y nodejs && \
+ARG NODE_VERSION
+RUN mkdir /nodejs && curl https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-linux-x64.tar.gz | tar xvzf - -C /nodejs --strip-components=1
+ENV PATH $PATH:/nodejs/bin
 
-    # Install Yarn
-    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+# Install Yarn
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
```
Philmod: Is the Yarn version important?

palmerj3: I think so. But I'm a little conflicted how to handle it given that the npm cloud builder is tagged based on the node version and not the NPM version.

palmerj3: That said - I think most people use the latest yarn stable and would care mostly about the node version.

Philmod: I agree with you on that.

If so, could we have the same logic as for the npm builder which builds many versions of nodejs? https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/npm/cloudbuild.yaml#L5

palmerj3: Yeah sounds good - I'll amend this PR with that.

Philmod: Awesome, thank you so much @palmerj3

palmerj3: @Philmod I just updated the PR. I tested everything locally.

PR #260, opened 2018-04-24
Add gcs-uploader image

ImJasonH:
This image incrementally uploads the contents of a directory to GCS and produces a source manifest (also in GCS) listing files that were uploaded or were already present.

Also:
- remove unnecessary vendored files in update-deps.sh
- remove dependency on glog (prefer log instead)
- build and tag docker images locally (then push) in cloudbuild.yaml, instead of pushing then pulling
- move gcs-fetcher to cmd/gcs-fetcher
- make default value `--help=false`, whoooops

Remaining work:
- add a sample/test and documentation in README.md


gcs-fetcher/cloudbuild.yaml
```diff
@@ -1,20 +1,36 @@
...
-- name: 'gcr.io/cloud-builders/docker'
-  args: ['pull', 'gcr.io/$PROJECT_ID/gcs-fetcher']
+# Push the images.
+- name: 'gcr.io/cloud-builders/bazel'
+  args: ['run', '--spawn_strategy=standalone', '--define', 'project=$PROJECT_ID', '//cmd/gcs-fetcher:push']
+- name: 'gcr.io/cloud-builders/bazel'
+  args: ['run', '--spawn_strategy=standalone', '--define', 'project=$PROJECT_ID', '//cmd/gcs-uploader:push']
 
-images: ['gcr.io/$PROJECT_ID/gcs-fetcher']
+# These are no-ops, but they tie the build to the images they produced, since
```
bendory: "These are no-ops, but they signal GCB that this build produced these images, since Bazel-based Docker pushes are not otherwise tracked."

ImJasonH:
Removed the `container_push` rules entirely, they didn't add anything anyway.

PR #270, opened 2018-05-01
Add yarn-puppeteer cloud builder

palmerj3:
Adds yarn-puppeteer cloud builder.

Works exactly like the yarn cloud builder but is based on alpine-node and contains chromium and necessary dependencies for running puppeteer.

If accepted I can create a similar npm-puppeteer.

### DIFF
--- a/yarn-puppeteer/Dockerfile
+++ b/yarn-puppeteer/Dockerfile
@@ -1,0 +1,24 @@
+ARG NODE_VERSION
+FROM node:$NODE_VERSION-alpine
+
+# Installs latest Chromium (64) package.
+RUN apk update && apk upgrade && \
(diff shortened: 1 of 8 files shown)


palmerj3:
@ImJasonH yeah I'd be happy to. I figured that might be suggested :)  My only thought was that google maintains puppeteer so it might be great to have a maintained cloud builder capable of using it since the yarn/npm ones are currently not.

ImJasonH:
That makes sense, unfortunately we don't work closely with that team at all, so unless they're willing to commit to fix breakages and keep the builder humming along smoothly, we'll probably have a better time in the community repo.

The good news is, it's _very_ easy to move a repo up from the community repo to the official one, so if it lives there and thrives we can revisit and move it.

PR #281, opened 2018-05-18
Unify CLIs for gcs-uploader and gcs-fetcher

ImJasonH:
With this change, `gcs-uploader` takes a `--location` flag, which must be the fully-specified object name of the manifest to upload to. This aligns with `gcs-fetcher`'s `--location` flag.

This change also:
- checks SHA1 digests if specified in the manifest
- shares the `ManifestItem` type between the fetcher and uploader packages


ImJasonH:
Looks like a couple of builds used the builder a couple weeks ago, I think we should be okay with a breaking change since this is largely an experimental builder that exists in service of testing the `gcs-fetcher` builder.

PR #216, opened 2018-03-01
Install all components in gcloud builder image

ImJasonH:

PR #286, opened 2018-05-29
Adds `gcr.io/cloud-builders/gcloud-slim`

ImJasonH:
This image skips installing all optional gcloud components, and as a result is much smaller:

```
$ docker images gcloud*
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
gcloud              latest              be897cf863a5        6 minutes ago       2.48GB
gcloud-slim         latest              7c2a4b1aece0        27 minutes ago      1GB
```

In environments where builder images might not be cached (e.g., local builder, etc.), having a smaller image can lead to faster builds.

For reference, `gcloud` started installing all optional components in #216 

/cc @dgageot


gcloud/cloudbuild.yaml
```diff
@@ -4,15 +4,21 @@
 steps:
 - name: 'gcr.io/cloud-builders/docker'
```
bendory: Instead of an Arg, would it make more sense to have two Dockerfiles in the source? Dockerfile.slim builds gcloud-slim, and Dockerfile.full starts with `FROM gcloud-slim` and adds components. WDYT?

ImJasonH: Say that's smart!

Unfortunately, in order to be `FROM gcr.io/cloud-builders/gcloud-slim` you have to first publish an image there. So for now this change builds two separate unrelated images, and once we've built `gcloud-slim` once we can change `gcloud` to build `FROM` it. WDYT?

ImJasonH: Nevermind, as soon as I wrote that I realized that was dumb and could do it a smarter way. PTAL!

gcloud/Dockerfile.slim
```diff
@@ -0,0 +1,43 @@
...
+    CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
+        --bash-completion=false \
+        --disable-installation-options && \
+
+    # TODO(jasonhall): These lines pin gcloud to a particular version.
+    # /builder/google-cloud-sdk/bin/gcloud components update --version 137.0.0 && \
+    # /builder/google-cloud-sdk/bin/gcloud config set component_manager/disable_update_check 1 && \
+    # /builder/google-cloud-sdk/bin/gcloud -q components update && \
+
+    # TODO(jasonhall): These lines pin gcloud to a particular version.
```
bendory: I'd like to drop these commented-out lines completely, but at the very least please only include them once.

PR #289, opened 2018-06-04
Add back bzip2

bendory:
Looks like bzip2 got accidentally deleted by #249.
This PR fixes regression reported https://github.com/GoogleCloudPlatform/cloud-builders/pull/197#issuecomment-394167764


bendory:
Actually, it looks like bzip2 is already installed:

> $ docker run --rm --entrypoint=/bin/bash gcr.io/cloud-builders/yarn:latest -- which bzip2
> /bin/bzip2
>

PR #299, opened 2018-07-02
Bazel container writes build IDs to output file.

michaeledgar:
The container now installs bazel and then installs a wrapper script that
calls bazel. The script determines if the user supplied the Build and/or
Invocation IDs by flag, and if not, adds flags with default IDs. Before
executing bazel, the IDs are written in JSON to the $BUILDER_OUTPUT/output
file.

### DIFF
--- a/bazel/bazel.sh
+++ b/bazel/bazel.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+# Always write the build_event_text_file to a temp file we create.
+readonly BUILD_EVENT_FILE="$(mktemp --tmpdir build_event_text_file.XXXXX)"
+
+/builder/bazel $@ --build_event_text_file="${BUILD_EVENT_FILE}"
+
+# Parse out the UUID from the BEP output file and write it to
+# $BUILDER_OUTPUT/output, whose first 4KB will be served inline in the Build.
+
+# 'uuid' is a field in the BuildStarted event that is guaranteed to appear first
+# in the build_event_text_file. That event might be larger than 4KB in text
+# format though. This grep solution is crude, but allows us to avoid actually
+# parsing the text proto file using real protobuf definitions.
+readonly INVOCATION_ID_FIELD="$(grep -Eo 'uuid: \".{36}\"$' "${BUILD_EVENT_FILE}" | head -n 1)"
+readonly INVOCATION_ID="${INVOCATION_ID_FIELD:7:-1}"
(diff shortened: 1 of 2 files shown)


bazel/cloudbuild.yaml
```diff
@@ -6,10 +6,12 @@ steps:
 - name: 'gcr.io/cloud-builders/docker'
   args: ['build', '--tag=gcr.io/$PROJECT_ID/bazel', '.']
 - name: 'gcr.io/$PROJECT_ID/bazel'
+  env: ['BUILD_ID=$BUILD_ID']
```
ImJasonH: I'm not super pumped that bazel users have to remember to pass this through to get Bazel to pick it up, but GCB's model currently doesn't support this by default either.

We've gotten FRs to always passthrough envs like `$BUILD_ID`, `$COMMIT_SHA`, etc., as env vars, maybe this is just more evidence we should consider it.

bazel/bazel.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+set -e
+
+# Bazel Wrapper script algorithm:
```
ImJasonH: Is there any chance we could convince Bazel to adopt this algorithm itself, and remove the need for this wrapper? Or is there something that Bazel already supports that we can piggyback off of? Injecting our own wrapper script isn't the worst thing in the world (we do it for `kubectl`) but I'd like to avoid it long-term if possible.

For `kubectl` we also retain the actual `kubectl` binary on the `PATH`, and only inject the wrapper in the builder's `ENTRYPOINT` -- this has the downside that users who do something like:

```yaml
- name: gcr.io/cloud-builders/kubectl
  entrypoint: bash
  args: ['-c', 'kubectl apply -f foo.yaml']
```

will bypass the wrapper, probably unknowingly. I'm not sure whether replacing the binary with the wrapper is more or less expected behavior.

bazel/cloudbuild.yaml
michaeledgar: For a general FR, I would say: If you passthrough $PROJECT_ID and $BUILD_ID, then steps can call GetBuild to fetch the rest everything else.

michaeledgar:
PTAL. I've come back around to this and updated the pull request. We now extract the invocation ID from a structured output file that Bazel writes, rather than intercepting the user's arguments.

Philmod:
I think this PR broke gcs-fetcher

PR #305, opened 2018-07-05
gcs-fetcher: Support fetching .tar.gz and .tar

dgageot:
Signed-off-by: David Gageot <david@gageot.net>

### DIFF
--- a/gcs-fetcher/pkg/fetcher/fetcher.go
+++ b/gcs-fetcher/pkg/fetcher/fetcher.go
@@ -16,7 +16,9 @@ limitations under the License.
 package fetcher
 
 import (
+	"archive/tar"
 	"archive/zip"
+	"compress/gzip"
 	"context"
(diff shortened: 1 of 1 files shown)


ImJasonH:
Do you have a need for this? StorageSource already supports fetching these types of archives, gcs-fetcher is mostly focused on fetching source manifests to enable incremental upload. Long-term I'd like to see if we can remove .tar(.gz) support entirely.

ImJasonH:
@squee1945 do you have an opinion about adding support for this to `gcs-fetcher`?

An alternative would be to migrate any existing users for archive support (are there any?) to a builder that only fetches archives (which could probably just be accomplished with a bash script using tar and unzip), and focus gcs-fetcher on fetching source manifests. WDYT?

squee1945:
If StorageSource supports this directly, why would we add it to gcs-fetcher?

PR #309, opened 2018-07-24
GCB rebrand.

bendory:


docker/cloudbuild.yaml
```diff
@@ -1,5 +1,5 @@
 # In this directory, run the following command to build this builder.
-# $ gcloud container builds submit . --config=cloudbuild.yaml
+# $ gcloud builds submit . --config=cloudbuild.yaml
```
ImJasonH: We can remove the `.` from this command too, and elsewhere.

Up to you.

bendory: I'm going to do that in a separate PR -- want to get this through for the rename...

Issue #322, opened 2018-08-06
RFC: Deprecate and turndown npm and yarn builder images


ImJasonH:
I would like to propose that we deprecate and stop building the `npm` and `yarn` builders, and instead recommend that users use the [official `node` image](https://hub.docker.com/_/node/) from Dockerhub.

This image gets regular updates from the official Node maintainers, and new versions of node are immediately available as builder images.

The only downside (AFAIK) is that the images don't specify an entrypoint, so instead of:

```
steps:
- name: gcr.io/cloud-builders/npm
  args: ['version']
```

users would have to write:

```
steps:
- name: node
  args: ['npm', 'version']
```

However, the `node` image includes both `yarn` and `npm`, and is updated and tagged with the version of node, so it's easy to specify in your build:

```
steps:
- name: node:6.14.3
  args: ['npm', 'version']
- name: node:10.8.0
  args: ['npm', 'version']
- name: node:10.8.0
  args: ['yarn', 'version']
```

I propose we recommend that users should use the official node image, include a deprecation warning in `gcr.io/cloud-builders/npm` and `/yarn`, and eventually break those images to force stragglers to use the official node image.

We would cache popular versions of the `node` builder on worker VMs, just as we do for `npm` and `yarn` builders today.

@palmerj3 @victorandree As recent contributors to the npm and yarn builders, I'm especially interested in your feedback.

P.S., if this work proves successful, we could possibly also migrate users of [`dotnet`](https://hub.docker.com/r/microsoft/dotnet/), [`mvn`](https://hub.docker.com/_/maven/), [`gradle`](https://hub.docker.com/_/gradle/), etc., to official images which receive updates more quickly.

victorandree:
I think this is a good idea!

It should make it easier to untangle `npm` from `node` releases for people who are e.g. stuck on node 8 but want to use a more recent version of `npm` (as I wanted to do in #263 and @nakleiderer recently wanted to). They can just add the extra `RUN npm install -g npm` layer to their images but still benefit from having the base image cached. Neat!

Personally, I would be very happy if this somehow meant that popular node images were immediately available as base images when doing Docker builds! I.e. that the `docker` cloud builder could share image cache, somehow. But maybe that's unrelated and too complicated.

ImJasonH:
This change wouldn't cache popular node packages, just the layers of the official node image, sorry.

victorandree:
I don't expect node packages to be cached!

[...]

PR #326, opened 2018-08-07
Base npm and yarn images on official node image

ImJasonH:
Related to #322 

There's been some discussion in the issue about whether to remove these images altogether, but in the meantime at least basing the images off the `node` image makes them simpler, and it means the `node` image will be cached on workers.

cc @palmerj3 @victorandree @ofrobots

### DIFF
--- a/npm/Dockerfile
+++ b/npm/Dockerfile
@@ -1,12 +1,3 @@
-# Use the base App Engine Docker image, based on debian jessie.
-FROM launcher.gcr.io/google/debian8
-
-# Install updates and dependencies
-RUN apt-get update -y && apt-get install --no-install-recommends -y -q curl python build-essential ssh git ca-certificates libkrb5-dev imagemagick && \
-    apt-get clean && rm /var/lib/apt/lists/*_*
-
-ARG NODE_VERSION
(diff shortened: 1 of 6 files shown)


ImJasonH:
FYI the `node` image is ~50% larger than the `debian8`-based images we used before.

```
$ docker images gcr.io/cloud-builders/npm
REPOSITORY                  TAG                 IMAGE ID            CREATED             SIZE
gcr.io/cloud-builders/npm   latest              31a73677aa30        About an hour ago   445MB

$ docker images node
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
node                latest              0e2811757f93        5 days ago          673MB
```

palmerj3:
@ImJasonH nice! Disregard my comment on the issue. Now I understand what you were going after. I thought you were suggesting to not have an entrypoint for npm/yarn. This is a very nice change!

ImJasonH:
@palmerj3 that was actually my proposal, and this is I think a good middle ground. Users who want versions of npm/yarn that we don't have builders for can use official images with entrypoint, and node will be cached as a base image, so I think everyone wins! 🎉

PR #331, opened 2018-08-09
Fix yarn / npm builder typo

codrienne:
cloudbuild.yaml will fail because of 'v10.3.0' isn't a valid version, but I assume '10.3.0' will work.


codrienne:
Testing locally...

PR #334, opened 2018-08-14
Update npm to latest node images

victorandree:
I went by the list on https://nodejs.org/en/download/releases/ updating:

* 6.14.0 -> ~6.14.3~ 6.14.4
* 8.11.0 -> ~8.11.3~ ~8.11.4~ 8.12.0
* 9.10.0 -> 9.11.2
* 10.3.0 -> ~10.8.0~ ~10.9.0~ 10.10.0

Update: Updated with [August 2018 security releases](https://nodejs.org/en/blog/vulnerability/august-2018-security-releases/).

The cloudbuild.yaml file also builds 8.4.0. I don't know why, so I didn't touch it.

### DIFF
--- a/npm/README.md
+++ b/npm/README.md
@@ -6,7 +6,7 @@ You might also consider using an [official `node` image](https://hub.docker.com/
 
 ```yaml
 steps:
-- name: node:10.9.0
+- name: node:10.10.0
   entrypoint: npm
   args: ['install']
 ```
--- a/npm/README.md
+++ b/npm/README.md
@@ -6,7 +6,7 @@ You might also consider using an [official `node` image](https://hub.docker.com/
 
 ```yaml
 steps:
-- name: node:10.8.0
+- name: node:10.9.0
   entrypoint: npm
   args: ['install']
 ```
--- a/npm/cloudbuild.yaml
+++ b/npm/cloudbuild.yaml
@@ -12,9 +12,9 @@ steps:
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
-  - '--build-arg=NODE_VERSION=8.11.3'
-  - '--tag=gcr.io/$PROJECT_ID/npm:node-8.11.3'
-  # 8.11.3 is tagged :latest
+  - '--build-arg=NODE_VERSION=8.11.4'
+  - '--tag=gcr.io/$PROJECT_ID/npm:node-8.11.4'
+  # 8.11.4 is tagged :latest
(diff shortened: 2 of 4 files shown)


merlinnot:
Great, my app breaks on previous versions of Node 10, thanks for this PR :)

victorandree:
@ImJasonH Some thoughts...

#### About supporting old versions
I think users of e.g. 10.3.0 are effectively abandoned by both the official node Docker images specifically, and the node project in general.

The [official images aren't being rebuilt for anything but the most recent LTS and current versions](https://hub.docker.com/r/library/node/tags/), as far as I can see. My impression is that node strictly follows semver for its releases, so it's unlikely that any less-than-current/LTS minor release would see a patch release (if you can update to the patch release, you can update to the new minor, right?).

This might be pretty node-specific so I can't say anything about other builders.

#### About tagged Docker images
To me, the point of using tagged Docker images is specifically to get an immutable image. So I would almost hope it didn't get rebuilt unnecessarily! It is perhaps then more unfortunate that no images got built for this (and the yarn) builder for 10.4.0, 10.4.1, 10.5.0, 10.6.0 and 10.7.0 node releases.

#### About what to do with these images
All in all, I agree that this adds to the case for deprecating this and the yarn image. I think caching popular base images on Cloud Build is _very_ nice (and my primary reason for submitting these PRs, to be fair 😄), but don't know what method you might want to follow to achieve that.

I personally would love to see e.g. all "recent" `node` and `python` base images cached and ready so I can get faster builds. This is getting a bit off-topic, but maybe you could consider caching images based on global usage, by mirroring https://hub.docker.com/explore/, or make it configurable per project.

#### About this PR specifically
Since I think the PR is useful, I can easily update it to keep building 10.3 (and any other minor versions), if that would help it be merged  🎉

[release page]: https://nodejs.org/en/download/releases/

victorandree:
@ImJasonH Any updates on this? 8.12.0 and 10.10.0 is out now.

ImJasonH:
@victorandree We can update this PR, or submit an updated version, if you'd like.

I still think the best long-term solution will be [asking users to just use the standard `node` image](https://github.com/GoogleCloudPlatform/cloud-builders/issues/322) at whatever version they want, and doing our best to keep up with caching the latest few versions on our workers.

victorandree:
@ImJasonH Agree completely about the way forward. I've pushed a commit to update 8.x to 8.12.0 and 10.x to 10.10.0

PR #339, opened 2018-08-16
Revert "Bazel container writes build IDs to output file."

Philmod:
Reverts GoogleCloudPlatform/cloud-builders#299

This broke gcs-fetcher builder: "Step #5: flag provided but not defined: -norun"
The --norun flag seems to be defined in rules_docker, described here: https://github.com/bazelbuild/rules_docker#using-with-docker-locally


Philmod:
Ping @michaeledgar

michaeledgar:
Confirmed that this rollback should land. The gcs-fetcher build failure is caused by incorrectly rewriting the bazel command line. It is *not* caused by misquoting, however.

In short: there is a bit of structure to the set of valid bazel command lines, setting the flag I wanted to set can't simply be prefixed or suffixed onto the configured command line. Fixing-forward will need a more carefully written wrapper script.

PR #344, opened 2018-08-22
Fix bazel.sh: insert bazel flag before any -- arg.

michaeledgar:
The bazel.sh wrapper script inserts a flag to control the file to which
structured output is written. This change ensures the flag is inserted
into the bazel command line where it will be parsed.

Previously the flag was naively appended to the bazel command line,
which broke [bazel run] invocations that pass arguments to the bazel-run
binary with the "--" separator.


michaeledgar:
Added an additional step to the bazel cloudbuild.yaml that passes arguments to the example binary. Changed the binary to print the arguments it receives, so we can (manually) verify that the arguments it receives are not mangled.

ImJasonH:
How about having the binary fail if its arg isn't exactly the arg you pass? Manual checking is for suckers.

ittaiz:
😍 which Bazel version does this mandate?

bazel/examples/main.go
```diff
@@ -1,7 +1,12 @@
...
 
-import "fmt"
+import (
+	"log"
+	"os"
+)
 
 func main() {
-	fmt.Println("Hello world")
+	if len(os.Args) != 2 || os.Args[1] != "world" {
```
ImJasonH: I made this change in your branch (something GitHub apparently lets me do for some reason!)

bazel/cloudbuild.yaml
```diff
@@ -12,10 +12,13 @@ steps:
...
   args: ['run', '--spawn_strategy=standalone', '//:target', '--verbose_failures']
   dir: 'examples'
+- name: 'gcr.io/$PROJECT_ID/bazel'
+  args: ['run', '--spawn_strategy=standalone', '//:target', '--verbose_failures', '--', 'world']
+  dir: 'examples'
 
 # Example was built as bazel:target.
 - name: 'gcr.io/cloud-builders/docker'
-  args: ['run', 'bazel:target']
+  args: ['run', 'bazel:target', 'world']
```
ImJasonH: Does this test need to include the `--` separator to test that the script is doing the right thing?

michaeledgar: Yes, but I also don't care about exercising the argument checking binary via docker. So I've since dropped this step entirely.

bazel/examples/main.go
michaeledgar: In parallel, I split this into two go_binaries: one that expects args, and one that doesn't care.

I also added additional arguments to the arg-checking step, to avoid bugs where our wrapper script only works for exactly 1 arg.

This ended up being a lot easier if I upgraded the go_* bazel rules to match gcs-fetcher's rules, so I did that too.

michaeledgar:
PTAL

ittaiz:
Is this immediately available on GCB? Is there a minimal bazel version?

ImJasonH:
This will be available after GCB's next release, probably Tuesday I would guess. I don't believe this requires any particular version of Bazel, but the Bazel image we build always uses the latest available version anyway.

PR #346, opened 2018-08-26
Send stderr to $BUILDER_OUTPUT/output.

squee1945:
Also, emit better message when AccessDenied trying to read bucket.

### DIFF
--- a/gcs-fetcher/pkg/fetcher/fetcher.go
+++ b/gcs-fetcher/pkg/fetcher/fetcher.go
@@ -23,10 +23,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"math"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -298,6 +317,17 @@ func (gf *Fetcher) fetchObjectOnce(ctx context.Context, j job, dest string, brea
 
 	r, err := gf.GCS.NewReader(ctx, j.bucket, j.object)
 	if err != nil {
+		// Check for AccessDenied failure here and return a useful error message on Stderr and exit immediately.
+		if strings.Contains(err.Error(), "<Code>AccessDenied</Code>") {
```
squee1945: I really dislike matching against the error message like this, but it seems like the structured error has already been unwrapped into a simple `error`, so the formal 403 code is gone. At least that's how it appeared to me.

squee1945:
Closing PR until I get the tests running under Bazel in Cloud Build.

PR #349, opened 2018-08-27
Emit actionable error message for access denied

squee1945:
Also, pipe all `stderr` to `$BUILDER_OUTPUT/output` if `$BUILDER_OUTPUT` is present in environment (also still emit to `stderr` regardless).

### DIFF
--- a/gcs-fetcher/cmd/gcs-fetcher/main.go
+++ b/gcs-fetcher/cmd/gcs-fetcher/main.go
@@ -66,9 +66,7 @@ func main() {
 		return
 	}
 
-	var stdout, stderr io.Writer
-	stdout = os.Stdout
-	stderr = os.Stderr
+	var stdout, stderr io.Writer = os.Stdout, os.Stderr
 	if outputDir, ok := os.LookupEnv("BUILDER_OUTPUT"); ok {
 		if err := os.MkdirAll(outputDir, os.ModePerm); err != nil {
 			logFatalf(os.Stderr, "Failed to create folder %s: %v", outputDir, err)
@@ -78,7 +76,11 @@ func main() {
 		if err != nil {
 			logFatalf(os.Stderr, "Cannot open output file %s: %v", outfile, err)
 		}
-		defer f.Close()
+		defer func() {
+			if cerr := f.Close(); cerr != nil {
+				log.Fatalf("Failed to close %q: %v", outfile, cerr)
+			}
+		}()
 		stderr = io.MultiWriter(stderr, f)
 	}
 
@@ -116,7 +118,7 @@ func main() {
 		Stderr:      stderr,
 	}
 	if err := gcs.Fetch(ctx); err != nil {
-		logFatalf(stderr, err.Error())
+		logFatalf(stderr, "failed to Fetch: %v", err.Error())
 	}
 }
 
--- a/gcs-fetcher/cmd/gcs-fetcher/main.go
+++ b/gcs-fetcher/cmd/gcs-fetcher/main.go
@@ -70,8 +70,7 @@ func main() {
 	stdout = os.Stdout
 	stderr = os.Stderr
 	if outputDir, ok := os.LookupEnv("BUILDER_OUTPUT"); ok {
-		err := os.MkdirAll(outputDir, os.ModePerm)
-		if err != nil {
+		if err := os.MkdirAll(outputDir, os.ModePerm); err != nil {
 			logFatalf(os.Stderr, "Failed to create folder %s: %v", outputDir, err)
 		}
 		outfile := filepath.Join(outputDir, "output")
--- a/gcs-fetcher/cmd/gcs-fetcher/main.go
+++ b/gcs-fetcher/cmd/gcs-fetcher/main.go
@@ -52,7 +52,7 @@ var (
 
 func logFatalf(writer io.Writer, format string, a ...interface{}) {
 	if _, err := fmt.Fprintf(writer, format+"\n", a...); err != nil {
-		log.Fatal("Failed to write log: %v", err)
+		log.Fatalf("Failed to write log: %v", err)
 	}
 	os.Exit(1)
 }
--- a/gcs-fetcher/cmd/gcs-fetcher/main.go
+++ b/gcs-fetcher/cmd/gcs-fetcher/main.go
@@ -50,6 +50,13 @@ var (
 	help        = flag.Bool("help", false, "If true, prints help text and exits.")
 )
 
+func logFatalf(writer io.Writer, format string, a ...interface{}) {
+	if _, err := fmt.Fprintf(writer, format+"\n", a...); err != nil {
+		log.Fatal("Failed to write log: %v", err)
+	}
+	os.Exit(1)
+}
+
 func main() {
 	flag.Parse()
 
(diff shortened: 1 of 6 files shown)


gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -298,6 +317,17 @@ func (gf *Fetcher) fetchObjectOnce(ctx context.Context, j job, dest string, brea
 
 	r, err := gf.GCS.NewReader(ctx, j.bucket, j.object)
 	if err != nil {
+		// Check for AccessDenied failure here and return a useful error message on Stderr and exit immediately.
```
ImJasonH: It seems like errors from `NewReader` should be `*googleapi.Error`s, is that not what you've seen in your tests?

https://godoc.org/cloud.google.com/go/storage#hdr-Errors

I thought you should be able to do:

```go
r, err := gf.GCS.NewReader(...)
if err != nil {
  if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == http.StatusForbidden {
    // got 403
  } else {
    // some other error
  }
}
```

If that's not what you're seeing though then maybe GCS has some other semantic difference between other GCP APIs.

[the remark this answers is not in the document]
squee1945: I've got this output to occur in a manual test - of course, I might have just won a buffering race. I also have a Go "error stack" that suggests this is where the error is raised.

squee1945: I tried casting to (googleapi.Error) which wouldn't compile. I didn't try (*googleapi.Error) though. I will give it a go.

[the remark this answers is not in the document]
squee1945: Okay, thanks to your help, I got this to work.

squee1945: Thanks to your help, I got this to work.

squee1945:
PTAL, I made changes to address your feedback.

gcs-fetcher/cmd/gcs-fetcher/main.go
```diff
@@ -59,19 +66,35 @@ func main() {
 		return
 	}
 
+	var stdout, stderr io.Writer
```
ImJasonH: Is there some reason you can't just do `stdout, stderr := os.Stdout, os.Stderr` (or some variation thereof)?

gcs-fetcher/cmd/gcs-fetcher/main.go
```diff
@@ -89,9 +112,11 @@ func main() {
 		Backoff:     *backoff,
 		SourceType:  *sourceType,
 		Verbose:     *verbose,
+		Stdout:      stdout,
+		Stderr:      stderr,
 	}
 	if err := gcs.Fetch(ctx); err != nil {
-		log.Fatal(err)
+		logFatalf(stderr, err.Error())
```
ImJasonH: How about `logFatalf(stderr, "Fetch: %v", err)` ?

gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -298,10 +320,25 @@ func (gf *Fetcher) fetchObjectOnce(ctx context.Context, j job, dest string, brea
...
+		if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == http.StatusForbidden {
+			// Try to parse out the robot name.
+			match := robotRegex.FindStringSubmatch(err.Error())
+			robot := "your Cloud Build robot"
```
ImJasonH: nit: "Cloud Build service account"

gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -298,10 +320,25 @@ func (gf *Fetcher) fetchObjectOnce(ctx context.Context, j job, dest string, brea
...
+			if len(match) == 2 {
+				robot = match[1]
+			}
+			gf.logErr("Access to bucket %q denied. You must grant Storage Object Viewer permission to %s.", j.bucket, robot)
+			os.Exit(1)
```
ImJasonH: It's pretty bad hygiene to have the `fetcher` library `os.Exit`, which shows in the ugliness in the test that has to jump through hoops to catch it. Can this just return an error that bubbles up to `main.go`'s call instead?

gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -317,7 +354,11 @@ func (gf *Fetcher) fetchObjectOnce(ctx context.Context, j job, dest string, brea
 		result.err = fmt.Errorf("creating destination file %q: %v", dest, err)
 		return result
 	}
-	defer f.Close()
+	defer func() {
+		if err := f.Close(); err != nil {
+			log.Fatalf("Failed to close file %q: %v", dest, err)
```
ImJasonH: Same with `log.Fatalf`s, would probably be better to have these errors returned by the method and handled in `main.go` instead.

gcs-fetcher/pkg/fetcher/fetcher.go
squee1945: The key reason for this P-R is to put a short, concise, actionable error message to the user if they get an AccessDenied. The way that Go "wraps" the error message makes this a bit tough. I think it would be worse to remove all the "wrapping context" from the errors.

By "wrapping context", I mean the way that a string is commonly prefixed on the error:
```
  err := something()
  if err != nil {
    return fmt.Errorf("failed to call something: %v", err)
  }
```

gcs-fetcher/pkg/fetcher/fetcher.go
squee1945: This too seems to be pretty tough in Go. Can you provide an example? Simply "return err" doesn't work well from a defer (that I can see).

gcs-fetcher/cmd/gcs-fetcher/main.go
squee1945: Sure, but you just proved my point below. :)

gcs-fetcher/cmd/gcs-fetcher/main.go
squee1945: They take the wrong type,  making the `stderr = io.MultiWriter(stderr, f)` assignment later on fail.

gcs-fetcher/pkg/fetcher/fetcher.go
ImJasonH: Yeah `defer` definitely makes this harder, unfortunately :disappointed: 

It's sometimes easier if you use a named return value:

```go
func foo(wc io.WriteCloser) (err error) {
  defer func() {
    cerr := wc.Close()
    if err == nil { err = cerr }
  }()
  if err := w.Write(...); err != nil {
    return err
  }
  return nil
}
```

https://play.golang.org/p/wAoWS8g_KbM

gcs-fetcher/pkg/fetcher/fetcher.go
ImJasonH: You don't have to wrap the error to return it, just return err and have code in `main.go` be responsible for `log.Fatal`ing or `os.Exit`ing.

gcs-fetcher/cmd/gcs-fetcher/main.go
ImJasonH: Oh, in that case, `var stdout, stderr io.Writer = os.Stdout, os.Stderr`

gcs-fetcher/pkg/fetcher/fetcher.go
squee1945: What I mean is that there are a lot of layers between this point and main and lots of additional wrapping strings are added along the way. For example:

- https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/gcs-fetcher/pkg/fetcher/fetcher.go#L227

But then to make matters worse, this method doesn't propagate the error and instead is collecting up a bunch of errors because there are up to 200 goroutines working in parallel.

Maybe I just take the performance hit and attempt to read from GCS early (e.g., in main.go) to capture this particular error?

ImJasonH: main.go doesn't have to be responsible for identifying the specific 403 error, it just has to be responsible for doing the os.Exit when it sees _any_ error returned by `Fetch`.

The work in `processJobs` could probably also be simplified by using an [`ErrGroup`](https://godoc.org/golang.org/x/sync/errgroup) which encapsulates cancelling all the other ongoing goroutines' contexts when one goroutine encounters an error, which could remove a lot of the layers of wrapping and propagating. It seems like that's what you're intending to do by `os.Exit`ing when you encounter the Access Denied, to stop all the other fetcher goroutines.

gcs-fetcher/cmd/gcs-fetcher/main.go
squee1945: Done.

gcs-fetcher/pkg/fetcher/fetcher.go
squee1945: Done.

gcs-fetcher/pkg/fetcher/fetcher.go
ImJasonH: No. If you'd prefer, we can merge and release this PR even with the `os.Exit` in the library, and I can do the work to refactor to use an `ErrGroup`. I'll make sure you get a chance to review and test that PR before we release it.

squee1945: ty

squee1945: In the refactor, the key thing is to get a simple error message written to $BUILDER_OUTPUT/output.

Good:
`Access to bucket XXX denied. You must grant Storage Object Viewer permission to YYY.`

Bad:
`failed to fetch: failed to fetchOnce: failed to fetchOnceWithTimeout: Access to bucket XXX denied. You must grant Storage Object Viewer permission to YYY.`

PR #352, opened 2018-08-28
GCS fetcher improvement grab-bag

ImJasonH:
- Use an `ErrGroup` to collect errors from worker goroutines
- Use a `context.WithTimeout` instead of the breaker chan
- Don't `os.Exit` from within the fetcher package, instead return an error that `main.go` can `log.Fatal`
- Improve unit testing to not fork another go test, since it won't `os.Exit` anymore.

/cc @squee1945

### DIFF
--- a/gcs-fetcher/Gopkg.lock
+++ b/gcs-fetcher/Gopkg.lock
@@ -2,6 +2,7 @@
 
 
 [[projects]]
+  digest = "1:6ab9ed0b0c5037e3c38b028f74a3986deef0003a55810c6261923e12eef76171"
   name = "cloud.google.com/go"
   packages = [
     "compute/metadata",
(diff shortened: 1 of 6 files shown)


squee1945:
Is there a tag other than `:latest` that we should be using in our build pipeline?

ImJasonH:
When GCB releases, it bakes this image in, and tags `:latest` to exactly the image by digest that was built during the release. If future processes push a `gcs-fetcher:latest` image, it will not start being used in GCB builds right away. Only when GCB builds the image with the changes during its release will it pick up the changes.

Contrast this with other images GCB caches, but does not itself release and manually tag to `:latest`, such as `ubuntu` -- when `ubuntu` gets pushed, GCB worker VMs will ~immediately pull that new image and make it available to builds. 

Does that answer your question?

ImJasonH:
We can build and push the image to any registry we want, with any tag. If you'd prefer us to push to `:canary` during our release that's definitely something we can do, and it would be immediately available to worker VMs as soon as new images are pushed to that tag.

Otherwise, we can also push it to another registry entirely for manual or automatic testing.

PR #363, opened 2018-09-17
FIX: Write invocation_id even when bazel fails.

michaeledgar:
Adds test script that causes bazel to produce a nonzero exit code and
verifies that the exit code is preserved while writing the invocation_id
to $BUILDER_OUTPUT/output.

### DIFF
--- a/bazel/bazel.sh
+++ b/bazel/bazel.sh
@@ -23,7 +23,13 @@ n=$(($#-$i+1))
 # Insert our flag at index $i out of $n. Quoting is all required for expansion.
 set -- "${@:1:$((i-1))}" "$BUILD_EVENT_FILE_FLAG" "${@:$i:$n}"
 
-/builder/bazel "$@"
+# Run bazel with the new command line and capture the exit code.
+#
+# This script has -e enabled, so if bazel exits non-zero, we need to both
+# capture the exit code *and* replace it with zero. That is the purpose of the
+# short-circuiting OR operator.
+EXIT_CODE=0
+/builder/bazel "$@" || EXIT_CODE=$?
 
 # Parse out the UUID from the BEP output file and write it to
 # $BUILDER_OUTPUT/output, whose first 4KB will be served inline in the Build.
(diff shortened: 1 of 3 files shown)


bazel/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM launcher.gcr.io/google/ubuntu16_04
 
+ADD invocation_id_test.sh /builder/invocation_id_test.sh
```
ImJasonH: Why is this added to the `bazel` image?

michaeledgar: I started by cargo-culting and forgot to switch to taking the test script from the build source. Fixed.

bazel/bazel.sh
```diff
@@ -23,7 +23,8 @@ n=$(($#-$i+1))
 # Insert our flag at index $i out of $n. Quoting is all required for expansion.
 set -- "${@:1:$((i-1))}" "$BUILD_EVENT_FILE_FLAG" "${@:$i:$n}"
 
-/builder/bazel "$@"
```
ImJasonH: Why not just

```
/builder/bazel "$@"
EXIT_CODE=$?
```

(Sorry if there's some obscure bashism that makes the `&&` and `||` preferable, it could use a comment if so)

michaeledgar: Good question, the reason why is subtle. The whole script is run with `set -e` at the top, which ensures that if *any* pipeline in the script fails, the whole script fails. (I actually picked this idiom up from reading Docker tutorials - it's not totally obscure). It's a safety measure for scripts where one can assume any failure is a total failure.

In a script with the `-e` option enabled, there are two ways to deal with a single pipeline that might fail in the middle of your script. 

1. Temporarily disable the option:
```bash
set +e
/builder/bazel "$@"
EXIT_CODE=$?
set -e
```
1. Change the pipeline to succeed, minimally by adding a short-circuiting logical or expression (`|| true`):
```bash
EXIT_CODE=0
/builder/bazel "$@" || EXIT_CODE=$?
```

I added a comment and switched to the last code snippet I've included here, though I'm fine with any formulation. I can even drop the `set -e` entirely, though you might not like the extra error checking that change would introduce.

ImJasonH: Thanks that makes perfect sense.

PR #364, opened 2018-09-17
Add gcloud to Gradle

scottefein:
To run gradle build appEngineDeploy, you need to have the gcloud command installed in the container. This adds that, so you can now deploy to App Engine Std from Cloud Build.

If you'd rather this go to community or if I should implement this differently, let me know!

### DIFF
--- a/gradle/Dockerfile
+++ b/gradle/Dockerfile
@@ -9,8 +9,34 @@ ARG BASE_URL=https://services.gradle.org/distributions
 ENV GRADLE_HOME "/usr/share/gradle-${GRADLE_VERSION}"
 ENV GRADLE_USER_HOME "${USER_HOME_DIR}/.gradle/"
 
-RUN apt-get update -qqy && apt-get install -qqy curl \
-  && mkdir -p /usr/share "${GRADLE_USER_HOME}" \
+RUN apt-get -y update && \
+    apt-get -y install gcc python2.7 python-dev python-setuptools wget ca-certificates default-jre curl && \
+
+    # Setup Google Cloud SDK (latest)
+    mkdir -p /builder && \
+    wget -qO- https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz | tar zxv -C /builder && \
+    CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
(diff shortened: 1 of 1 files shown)


aslo:
cc @meltsufin to review

meltsufin:
cc @dlorenc

ImJasonH:
I have a couple concerns with this change, which I'd like to discuss at least:

1.) It might be easier to make this `FROM gcr.io/cloud-builders/gcloud-slim`, or to make the `javac` image `FROM` it. That way we only have to look in one place to debug gcloud installation issues, etc.

2.) The `gcloud-slim` image is still, despite its name, ~1 GB in size, which makes the `gradle` builder that much bigger as a result. Both images will be cached when used in GCB, but won't necessarily be in other execution environments. Not all `gradle` users would necessarily want `gcloud` installed, and all would pay the cost.

Would it make sense to create a new `gcr.io/cloud-builders/gradle-gcloud` image, that installs both tools (by being `FROM` one and installing the other)? This image would only be necessary for gradle users who want to use the `appEngineDeploy` feature, and wouldn't need to be used by "vanilla" gradle users.

lesv:
This might address #418

lesv:
It probably will work now if your build environment isn't gVisor.  If it is, you'll probably still need it.

ImJasonH:
@lesv I'm confused how gVisor plays in, can you elaborate?

@nlopezgi has also investigated some work to produce images (on-demand?) that include separate packages like gradle and gcloud, so we don't have to build them bespoke in each case. This issue is a good use case for that.

lesv:
It depends on where you install to. If you install gcloud to a mounted writable disk, then all is well.  If you attempt to install to the main image in Docker,it's write protected in GVisor, and it will fail.  (Believe me, I've tried)

nlopezgi:
thanks for looping me in @ImJasonH ,
we've run into some technical difficulties, we're regrouping to see if it will be possible to provide such images on demand, but likely we wont have something coming out in the short term

scottefein:
Any update on this? We're still maintaining this externally for launching to GAE.

bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/665 and https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #366, opened 2018-09-18
gcloud: add make

jayceekay:
let users reuse makefiles which invoke gcloud commands

### DIFF
--- a/gcloud/Dockerfile
+++ b/gcloud/Dockerfile
@@ -22,6 +22,7 @@ RUN apt-get -y update && \
         emulator-reverse-proxy \
         kubectl \
         pubsub-emulator \
+        make \
         && \
 
     /builder/google-cloud-sdk/bin/gcloud -q components update && \


ImJasonH:
I'm a bit :-1: on adding tools to this (already massive) builder, just in case people want to use them. Why not create a `make` builder that has `gcloud` installed instead? Or build your own image that's `FROM gcr.io/cloud-builders/gcloud` (or `gcloud-slim`) and installs `make`.

listx:
I think supporting Makefiles is reasonable, only because they are so ubiquitous in the computing world.

jayceekay:
I noticed some of the cloud builders had them (eg the docker container I think) but that this one didn’t, so figured I’d add it since I’m sure I’m not the only one who wants the commands to be a little more portable/consolidated since they’re not specific to just cloud build

directionless:
I would find this useful.

I have a mid-level complexity makefile, and I have a cloudbuild.yaml, and there's a bunch of tension between the steps. Right now, my various cloudbuild steps call things like `make build` and `make deploy` using appropriate containers. It would be great for my cloudbuild `make deploy` to be able to use the pre-existing `gcr.io/cloud-builders/gcloud` and save me the trouble of making my own.

But, I can't help but feel like I have a lot of duplication between make and cloudbuilder there, and this need, perhaps, indicates some other process issue

ImJasonH:
I don't think we intend to add Make to the `gcloud` image, since we otherwise hope to move to use the official `google/cloud-sdk` image directly, which the Cloud SDK team maintains themselves.

If you find that you need `make` and `gcloud` in the same image, you might consider creating a custom builder image that is `FROM google/cloud-sdk` and which installs Make.

listx:
Just adding some more metadata for posterity:

This is the page for google/cloud-sdk: https://hub.docker.com/r/google/cloud-sdk/

From grokking the announcement mailing list https://groups.google.com/forum/#!forum/google-cloud-sdk-announce, I found https://b.corp.google.com/issues/new?component=187143&template=800102. @jayceekay feel free to migrate this issue there.

Issue #367, opened 2018-09-19
docker: enable experimental commands


listx:
The motivation behind this is from a Kubernetes project (https://github.com/kubernetes/dns/commit/50149718d9859722aa1db4af2f4137ab74fa82d1) that would like to use the experimental `docker manifest` command. Currently we use version 17.12.0-ce of docker, which does not support this. According to https://docs.docker.com/release-notes/docker-ce/, the support for `docker manifest` was added in the 18.02 release.

The logic in https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/docker looks pretty straightforward for building the 3 different 17.x versions we have today, so I'll try creating a patch there.

prameshj:
We need docker 18.06+ in order to build the images correctly - https://github.com/kubernetes/kubernetes/pull/68430#issuecomment-422991629

listx:
As @prameshj said, we should support 18.06+ of docker CLI in order to build "fat" manifests (with the experimental `docker manifest` command) correctly.

There still remains a question though in how users of a hypothetical 18.06+ image would use it to run experimental commands.

Option 1: Experimental commands are disabled by default, and users have to write the config at runtime like this:
```
root@65a4552dc80f:~# mkdir .docker
root@65a4552dc80f:~# echo '{"experimental":"enabled"}' > ~/.docker/config.json
```
in order to call `docker manifest`.

Option 2: We have the above setting injected already when we build the 18.06+ image, so that experimental commands are always enabled. I would vote for having it enabled by default because, ultimately it's still up to the user if they want to use experimental subcommands. So, there's no "hidden action from a distance" magic happening here.

codrienne:
I've noted your request to support Docker 18.06+ and enable experimental commands by default, though I'm not sure we should since they are [not ready for production](https://github.com/docker/docker-ce/tree/master/components/cli/experimental). At the very least, it would be helpful to document how to enable it.

listx:
Going with Option 1 is fine too. Ultimately the difference between Option 1 vs 2 will go away on its own when docker manifests become first-class citizens (a year from now? 2 years?).

listx:
Hmm actually to clarify, I suppose there will always be experimental Docker commands as other new ones pop up in the future.

In this case I think going with Option 1 is much saner. Where would we document the enabling of the experimental flag @codrienne ?

prameshj:
[...]

PR #368, opened 2018-09-19
docker: add support for 18.03.1

listx:
Also tag this version as "latest".

I did a test with the 18.03.1 image built from this change, and apparently enabling the experimental "manifest" command is doable by just modifying a config file, as below:

```
root@65a4552dc80f:~# docker manifest
docker manifest is only supported when experimental cli features are enabled
root@65a4552dc80f:~# ls
root@65a4552dc80f:~# mkdir .docker
root@65a4552dc80f:~# echo '{"experimental":"enabled"}' > ~/.docker/config.json
root@65a4552dc80f:~# docker manifest

Usage:	docker manifest COMMAND

Manage Docker image manifests and manifest lists

Options:


Commands:
  annotate    Add additional information to a local image manifest
  create      Create a local manifest list for annotating and pushing to a registry
  inspect     Display an image manifest, or manifest list
  push        Push a manifest list to a repository

Run 'docker manifest COMMAND --help' for more information on a command.
root@65a4552dc80f:~# exit
exit
```

I wonder if we should do this by default in `Dockerfile-18.03.1` itself? Or should we tell users to run the 3-command incantation if they want to enable experimental commands (like `docker manifest`, and possibly others in the future). I would vote for having it enabled by default because, ultimately it's still up to the user if they want to use experimental subcommands. Thoughts?

### DIFF
--- a/docker/Dockerfile-18.03.1
+++ b/docker/Dockerfile-18.03.1
@@ -1,0 +1,20 @@
+FROM launcher.gcr.io/google/ubuntu16_04
+
+RUN apt-get -y update && \
+    apt-get -y install \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        make \
+        software-properties-common && \
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
+    apt-key fingerprint 0EBFCD88 && \
+    add-apt-repository \
+       "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+       xenial \
+       edge" && \
+    apt-get -y update && \
+    apt-get -y install docker-ce=18.03.1~ce-0~ubuntu
+
+ENTRYPOINT ["/usr/bin/docker"]
+
--- a/docker/cloudbuild.yaml
+++ b/docker/cloudbuild.yaml
@@ -24,6 +24,13 @@ steps:
   - '--file=Dockerfile-17.12.0'
   - '.'
   id: '17.12.0'
+- name: 'gcr.io/cloud-builders/docker'
+  args:
+  - 'build'
+  - '--tag=gcr.io/$PROJECT_ID/docker:18.03.1'
+  - '--file=Dockerfile-18.03.1'
+  - '.'
+  id: '18.03.1'
 # Future supported versions of docker builder go here.
 
 # Test each version by using it to run "docker build" on itself.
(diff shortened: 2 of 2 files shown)


codrienne:
Thanks for the PR! I'm afraid I'm going to have to close it though :\

Supporting a new docker version requires multiple steps / internal testing by the Cloud Build team, and is something we have to do ourselves.

Note about docker manifest: it's been brought to my attention that it is broken for 18.03. See https://github.com/docker/cli/issues/1135. The manifest list built with `docker manifest` is wrong. 
To quote a colleague: "Docker can still pull the wrong images, because it doesn't perform proper size validation. However, containerd can't pull those images, because it validates the image correctly. The issue was fixed in 18.06-ce, but not cherry-picked into 18.03-ce."  When we do a Docker upgrade we will make sure to use 18.06+.

Could you open a separate issue with your note about experimental commands?

prameshj:
Yes, we need docker 18.06+ in order to build images correctly. https://github.com/kubernetes/kubernetes/pull/68430#issuecomment-422991629

ImJasonH:
If I understand the issue correctly, malformed multi-arch images are built by docker versions between `18.03` and `18.06`, which is why it's important to update past `18.06+`.

If that's the case, then I don't think GCB is involved at all, since our docker version is currently `17.12.0`, which doesn't support `docker manifest` at all.

Were malformed images produced by GCB builds?

listx:
> Thanks for the PR! I'm afraid I'm going to have to close it though :\
> 
> Supporting a new docker version requires multiple steps / internal testing by the Cloud Build team, and is something we have to do ourselves.
> 
> Note about docker manifest: it's been brought to my attention that it is broken for 18.03. See [docker/cli#1135](https://github.com/docker/cli/issues/1135). The manifest list built with `docker manifest` is wrong.
> To quote a colleague: "Docker can still pull the wrong images, because it doesn't perform proper size validation. However, containerd can't pull those images, because it validates the image correctly. The issue was fixed in 18.06-ce, but not cherry-picked into 18.03-ce." When we do a Docker upgrade we will make sure to use 18.06+.
> 
> Could you open a separate issue with your note about experimental commands? I think it's worth discussion.

listx:
@codrienne I retitled the issue I opened earlier https://github.com/GoogleCloudPlatform/cloud-builders/issues/367

@ImJasonH As you stated, the docker:latest in this repo is based on 17.12.0 and so those comments about 18.03 being tainted does not apply --- unless, of course, someone went ahead (like I did in this PR) and created their own 18.03 and build (manifest) images that way.

ImJasonH:
@listx Even if you built your own docker builder than used 18.03 of the client, builds wouldn't work because the daemon it attaches to on the worker node would be 17.12, and the client isn't allowed to be newer than its daemon.

listx:
@ImJasonH Ah I missed that bit about cli vs daemon versions; good point.

PR #375, opened 2018-09-28
remove gradle.properties from $GRADLE_USER_HOME

leo-baltus:
as it overrules gradle.properties in project root directory
https://docs.gradle.org/current/userguide/build_environment.html

### DIFF
--- a/gradle/Dockerfile
+++ b/gradle/Dockerfile
@@ -20,7 +20,5 @@ RUN apt-get update -qqy && apt-get install -qqy curl \
   && apt-get remove -qqy --purge curl \
   && rm /var/lib/apt/lists/*_*
 
-ADD gradle.properties "${GRADLE_USER_HOME}"
-
 ENTRYPOINT ["/usr/bin/gradle"]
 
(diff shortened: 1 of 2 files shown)


leo-baltus:
I signed it!

meltsufin:
Shouldn't the `gradle.properties` from Gradle home and the project root be effectively merged by Gradle?
I imagine that there was a reason for `org.gradle.daemon=false` in the first place.

leo-baltus:
Well, the documentation (see url above) says:
> The configuration is applied in following order (if an option is configured in multiple locations the last one wins)

So, no merging. Would be a good idea though :-)

meltsufin:
This file in question seems to be only configuring `org.gradle.daemon=false`. Are you trying to override this particular property?
cc/ @loosebazooka

leo-baltus:
No, I have all sort of settings I want to have applied. Mostly var=value

leo-baltus:
It seems I have mistaken an error I had with properties being not set and have misread the documentation.
Sorry for the noise.

PR #378, opened 2018-10-10
 Place manifest and zip archive in staging folder.

squee1945:
The manifest and zip archive are currently polluting the user's namespace. This change places those temporary files into the staging folder, which is scrubbed out after completion.


gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -475,19 +482,21 @@ func (gf *Fetcher) fetchFromManifest(ctx context.Context) error {
 	log.Printf("Fetching manifest %s.", formatGCSName(gf.Bucket, gf.Object, gf.Generation))
 
 	// Download the manifest file from GCS.
+	manifestDir := gf.StagingDir
```
squee1945: Maybe `ioutil.TempDir("", "gcs-fetcher")` is better than `gf.StagingDir`?

PR #390, opened 2018-10-17
Run go mod vendor and update update-deps.sh

ImJasonH:
This updates the cached deps in `vendor/` and update `update-deps.sh` to ensure this is done when new deps are updated.

Also fix a lint error in `cmd/gcs-uploader/main.go`

cc @dprotaso


gcs-fetcher/cmd/gcs-uploader/main.go
```diff
@@ -54,7 +54,7 @@ func main() {
 	}
 	bucket, object, generation, err := common.ParseBucketObject(*location)
 	if err != nil {
-		log.Fatalln("parsing location from %q: %v", *location, err)
+		log.Fatalf("parsing location from %q: %v", *location, err)
```
dprotaso: \n at the end

ImJasonH: `log` methods already append a newline at the end. I'm not sure why both `log.Print` and `log.Println` exist...

dprotaso: cool

dprotaso: lgtm

dprotaso:
@codrienne it's good to merge without the `\n` change.

PR #392, opened 2018-10-17
RFC: Make gcr.io/cloud-builders/dotnet FROM microsoft/dotnet:sdk

ImJasonH:
This is part of our ongoing effort to recommend that users use official container images maintained by the team that builds the underlying tool. Previously, we've made `gcr.io/cloud-builders/npm` simply be `FROM nodejs; ENTRYPOINT ["npm"]`, etc. (https://github.com/GoogleCloudPlatform/cloud-builders/pull/326)

This removes some capabilities from the dotnet builder image, since it currently installs 3 versions of the .NET SDK and selects among them based on project config. Now, this image simply provides the single most recent SDK version. If users want to use an older SDK version, they can do so by running an image tagged with that version, e.g.:

```yaml
steps:
- name: microsoft/dotnet:1.1-sdk
  entrypoint: dotnet
  args: ['build']
```

Please let me know if this breaks any of your tooling or processes. In general I think users will have a better time relying on the official images than having to wait for us to update our SDK version.


cbuchacher:
So, if I understand correctly the cloud builder will provide only one version, which is the most recent version. What is the definition of most recent version? Is it the most recent production version? Or will the 2.2 preview be considered the most recent version?

In general, I would prefer having fixed versions, e.g. gcr.io/cloud-builders/dotnet:2.1.500. I don't mind older versions disappearing, but it is a frequent inconvenience and cause of confusion if the behavior suddenly changes because the dotnet version happens to change underneath.

Obviously I could use the official images, but if possible I would like to use the cached cloud-builder images to avoid downloading the official image on every build (which can take between 1-3 minutes).

Finally, I wonder how will I tell which version of the dotnet sdk the cloud builder is based on currently? For example, how can I check at some point in time if gcr.io/cloud-builders/dotnet is at 2.1.302 or at 2.1.500?

ImJasonH:
@cbuchacher if it's caching you're concerned about, we can cache some small number of the most popular official `dotnet` builders.

To check what version is installed, you can include a step like this in your build:

```
- name: gcr.io/cloud-builders/dotnet
  args: ['version']
```

When that image is based on, for example, `dotnet:2.1-sdk` then it will automatically increment to new point versions, but presumably never go past a `2.1` minor version release.

cbuchacher:
If you could cache the most common tags, i.e. `microsoft/dotnet:2.1-sdk`, `microsoft/dotnet:2.1-aspnetcore-runtime` and `microsoft/dotnet:2.1-runtime`, that would be awesome. In fact, I think it would be better than having a duplicate cloud-builder image which is more or less up-to-date.

ImJasonH:
I've proposed a change internally to cache `microsoft/dotnet:1.1-sdk`, `microsoft/dotnet:2.1-sdk` (which was already cached as `microsoft/dotnet:latest`) and `microsoft/dotnet:2.2-sdk`. I can't comment on whether/when that will reach production. I know basically nothing about .Net, can you explain why you'd need the `runtime` images cached?

Note that we have limited cache space, and at this time any additional cached images eat into the disk space available to users' builds, so we try to keep it as small as possible and only populated with widely-used images where they can provide the most benefit.

cbuchacher:
Most Dockerfiles for dotnet projects have two stages as such:

```Dockerfile
FROM microsoft/dotnet:2.1-sdk AS base
WORKDIR /build
COPY . .
RUN dotnet publish

FROM microsoft/dotnet:2.1-aspnetcore-runtime
WORKDIR /app
ENTRYPOINT ["dotnet", "App.dll"]
COPY --from=base /build/bin/publish/ .
```

The reason is that the sdk image is very large and is only need to build the dotnet project. The runtime image on the other hand is much smaller. In the second stage, only the build output from the first stage is included and the runtime image contains the runtime dependencies. The build time dependencies from the sdk image are discarded.

Therefore, the sdk image is usually used together with a runtime image. According to https://hub.docker.com/r/microsoft/dotnet/, those are the most commonly used images:

> The following tags are the latest stable versions of the most commonly used images. The complete set of tags is listed further down.
>
> 2.1-sdk
> 2.1-aspnetcore-runtime
> 2.1-runtime

cbuchacher:
Here are the current image sizes:
```
microsoft/dotnet     2.1-sdk                      6a9b6788c55f        1.73GB
microsoft/dotnet     2.1-aspnetcore-runtime       40d759655ea3        255MB
microsoft/dotnet     2.1-runtime                  e0ac7f65cd5c        180MB
```

ImJasonH:
Thanks, that's really helpful. So we'd get the most benefit from caching `sdk` images, and somewhat less from caching `runtime` images. We are effectively only caching an SDK image today, in `gcr.io/cloud-builders/dotnet`.

I think we'll probably continue with just caching `sdk` images for now, and assess later whether it's worth caching the smaller `runtime` images -- you should still get the benefit of not having to pull that large image, though you'll still have to pull the smaller `runtime` image.

ImJasonH:
There hasn't been any feedback on this PR for months, and the dotnet image is currently broken due to changes in debian8 which it depends on. I'm going to merge this PR to unwedge releases, if anybody has any concerns after this rolls out I will personally help unbreak you.

PR #393, opened 2018-10-18
include the license for gcs-* cmds in the container image

dprotaso:
I copied the LICENSE to the sub-folder since docker build doesn't like symlinks outside the build context


dprotaso:
oh yeah - no prow lol

Issue #404, opened 2018-11-05
Go: Unable to use PROJECT_ROOT if /src directory exists


mattayes:
Given the following (trimmed-down) repo layout:

* github.com/ORG/ROOT
  * cloudbuild.yaml
  * src/
    * ...
  * vendor/
    * ...

Given this setup, we need to set PROJECT_ROOT to `github.com/ORG/ROOT` to create the `./gopath` directory. Unfortunately, since our repo also has a `src` directory, `PROJECT_ROOT` is ignored and `$PWD` is used as `GOPATH`:

https://github.com/GoogleCloudPlatform/cloud-builders/blob/bcbc1850da7df3dcf7f42de8673ca1d8a67354b6/go/prepare_workspace.inc#L33-L37

`PROJECT_ROOT` should be checked before the `src` directory.

PR #405, opened 2018-11-05
Go: check for PROJECT_ROOT before src when setting GOPATH (#404)

mattayes:
Issue: #404 

Changes the order of operations when setting GOPATH.

Former order:

1. Check `$GOPATH`.
1. Check `./src`.
1. Check `$PROJECT_ROOT`.
1. Check `.go` files.

New order:

1. Check `$GOPATH`.
1. Check `$PROJECT_ROOT`.
1. Check `./src`.
1. Check `.go` files.

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -31,13 +31,12 @@ Before the `go` tool is used, the build step first sets up a workspace.
 To determine the workspace structure, this tool checks the following, in order:
 
 1.  Is `$GOPATH` set? Use that.
-2.  Is there a `./src` directory? Set `GOPATH=$PWD`.
-3.  Is `$PROJECT_ROOT` set? Make a temporary workspace in `GOPATH=./gopath`, and
+2.  Is `$PROJECT_ROOT` set? Make a temporary workspace in `GOPATH=./gopath`, and
     link the contents of the current directory into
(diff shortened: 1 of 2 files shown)


mattayes:
@ImJasonH I noticed you're one of the more active [contributors](https://github.com/GoogleCloudPlatform/cloud-builders/graphs/contributors). Do you know who should review this PR?

mattayes:
Thanks @Philmod, @skelterjohn, @ImJasonH!

skelterjohn:
a bit curious - when do you have an src directory where PROJECT_ROOT is appropriate? It's possible, and your PR seems like the right way to handle it (PROJECT_ROOT is an affirmative choice and takes precedence over implicit config), but how did it come up? :)

mattayes:
Here's our directory structure:

```
$GOPATH/src/github.com/<org>/<project>/src/...
```

Our import paths look like this:

```
// example.go
package main

import github.com/<org>/<project>/src/...
```

skelterjohn:
That extra `src` there will do it, for sure. I'll note that this is not conforming to typical Go conventions but there is no reason that the builder can't deal with it properly.

PR #408, opened 2018-11-20
Send non-kubectl message to stderr, exec kubectl

discordianfish:
Took me way to long to figure out why I couldn't parse the `kubectl -o json` that came from the wrapper until I realize it prints the invocation line.


discordianfish:
I'm pretty sure I signed that already.

ImJasonH:
#456 was just merged which conflicts with this change. Should be a fairly straightforward merge though.

I don't see a signed CLA in our system, so you'll need to sign that before we can accept this change.

discordianfish:
@ImJasonH resolved conflict

PR #411, opened 2018-11-28
feat(npm): update node to latest version

AdriVanHoudt:
This includes security releases.

Let me know if I missed something since first PR.

### DIFF
--- a/npm/cloudbuild.yaml
+++ b/npm/cloudbuild.yaml
@@ -6,15 +6,15 @@ steps:
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
-  - '--build-arg=NODE_VERSION=6.14.4'
-  - '--tag=gcr.io/$PROJECT_ID/npm:node-6.14.4'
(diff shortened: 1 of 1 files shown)


ImJasonH:
This change LGTM, but we should more forcefully recommend people use the official node/yarn images directly, so these kinds of PRs are not necessary.

AdriVanHoudt:
I guess that is on your end, if you intent to merge I can update those to the latest version before you do

ImJasonH:
Yeah, let's upgrade to the latest available versions in the hopes that they're the last we have to support this way.

#451 addresses more forcefully recommending official images for mvn/npm/yarn

ImJasonH:
We're going to recommend users run the official `node` image directly, which currently points to `11.9.0` and which is already cached on worker images.

Over time we can upgrade the `npm` and `yarn` builders to just be `FROM node` (latest) and remove a lot from these builders' configs.

AdriVanHoudt:
Cool, I will use that then, thanks!

PR #414, opened 2018-12-04
Fix + cleanup broken build issues.

bendory:


bendory:
Note: this PR fixes the breakage in the use of the native `git_repository` rule. build step still has a build failure in target `:go_context_data`.

bazel/cloudbuild.yaml
```diff
@@ -1,5 +1,5 @@
 # In this directory, run the following command to build this builder.
-# $ gcloud builds submit . --config=cloudbuild.yaml
+# $ gcloud builds submit
```
Philmod: no need for the source / config?

bazel/.gcloudignore
```diff
@@ -0,0 +1,3 @@
+cloudbuild.yaml
```
Philmod: why is that?

bazel/cloudbuild.yaml
bendory: No need -- the default values are no ``config=cloudbuild.yaml` and `.`

bazel/.gcloudignore
bendory: `gcloud` reads the `cloudbuild.yaml` and converts it into the build request; there is no need to submit it to the `/workspace` for the build.

Issue #417, opened 2018-12-10
Provide Gradle builder for Gradle v5.0 (and possibly Gradle v4.10.3)


dubinsky:
Latest Gradle version supported by the builder is v4.6.
Projects that use latest versions of certain plugins (e.g., `com.bmuschko.docker-remote-api`) or more current Gradle features (e.g., publishing) require later versions of Gradle.

Strat1987:
would also appreciate bumping the version for this builder. Also at this time it seems it uses the 4.0 as a default. Can you please also update the readme on how to use explicit version?

PR #437, opened 2019-01-21
added support for gradle 4.10.3, 5.0, 5.1.1 and javac 11

etsrepo:
Hi,

updated cloudbuild.yaml in gradle and javac to support gradle 4.10.3, 5.0, 5.1.1 and javac 11. 
I used openjdk:11 for javac 11 because there is no 'launcher.gcr.io/google/openjdk11' available.

example spring builds are successful with Gradle 5.x and javac 11

please let me know if anything I need to do anything to for this merge.

### DIFF
--- a/gradle/cloudbuild.yaml
+++ b/gradle/cloudbuild.yaml
@@ -8,6 +8,66 @@ steps:
   args:
   - 'build'
   - '--build-arg=BASE_IMAGE=gcr.io/$PROJECT_ID/javac:8'
+  - '--build-arg=GRADLE_VERSION=5.1.1'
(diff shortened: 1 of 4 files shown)


etsrepo:
fixes https://github.com/GoogleCloudPlatform/cloud-builders/issues/417

ImJasonH:
@loosebazooka you were nominated as a person on the Cloud Tools team who might be able to help us maintain the maven/gradle builders.

ImJasonH:
We recommend using the official `gradle` image directly. `gradle:latest`, which we will cache, currently points to `gradle:5.2.0-jdk8` according to https://hub.docker.com/_/gradle

PR #440, opened 2019-01-25
awscli image support

tonymet:
- based on `python3-alpine` . Added compatibility pkgs
- ~140mb in size
- see README for usage. 
- great for migrating projects out of AWS and into GCP


tonymet:
doh! pushed to the wrong repo. i'll move to the community

Issue #443, opened 2019-01-29
Feedback is poor


wyaeld:
Cloud Build is a great tool solving a huge problem, but it disappointing to see community contributions of various potential improvements or support languish for days/weeks with zero feedback from the team.

Not every idea or PR is worth of being merged, but they should all be treated as being worth of response.  Its a reasonable expectation that most of them come from paying customers of the platform, and that a silent majority probably looks and gets frustrated by the inaction rather than raising issues directly at you like this :-)

Its not really clear who to direct anything at, so I'll just ping @ImJasonH @nof20 and roll the dice.

nof20:
Hey @wyaeld , thanks for the feedback.  You're right there are both issues and PRs sitting around without feedback: as we say in the UK, it's a fair cop.

Are you referring to this repo, [cloud-builders-community](http://www.github.com/GoogleCloudPlatform/cloud-builders-community), or both?

@ImJasonH for his thoughts too.

wyaeld:
I was limiting it to this repo, so people can understand what is and isn't going to be officially supported or possible.

I'm in NZ, we use UK expressions too :-)

wyaeld:
appreciate it @ImJasonH 

Expect to be working on GCP for a long time.  Huge fan of what you guys are doing.

PR #445, opened 2019-02-01
Create issue template

ImJasonH:
This recommends filing issues in our public issue tracker instance rather than on the GitHub issue tracker

Hopefully begins to address issues raised in https://github.com/GoogleCloudPlatform/cloud-builders/issues/443


nof20:
Looks good, thanks.

Issue #412, opened 2018-11-28
New builders for Maven and Java 11


Alos:
It would be nice if we had new versions of gcr.io/cloud-builders/javac and gcr.io/cloud-builders/mvn with support for OpenJDK 11.

akauppi:
Connected to this, I wonder if they could be named other than `javac`.

- the current image also contains Docker (attributing to its huge 980 MB size for a JDK 8 image).
- other builders (like mine, for sbt+Scala) only need it for the JDK.

I only realized Docker was included when digging why the size of the image is so big. Having this info in the name would be helpful.

I'll likely name my derived builder `sbt-scala-docker` for this reason, to make the existence of Docker explicit. And make another `sbt-scala` without Docker, though... in my use cases it'll always be needed so maybe I don't bother.

btw. do the other GCR builder images also contain Docker, implicitly?

bendory:
https://github.com/GoogleCloudPlatform/cloud-builders/issues/665

PR #448, opened 2019-02-01
add support for maven 3.6.0

etsrepo:
Hi,

Added support for maven 3.6.0. 

changed sha256sum to sha512sum, because maven 3.6.0 distributing sha512 in official maven site. I have recalculated sha512sum for 3.5.0 and 3.3.9 and updated docker file accordingly.

Ref:
https://archive.apache.org/dist/maven/maven-3/3.6.0/binaries/apache-maven-3.6.0-bin.tar.gz.sha512
https://archive.apache.org/dist/maven/maven-3/3.6.0/binaries/

### DIFF
--- a/mvn/Dockerfile
+++ b/mvn/Dockerfile
@@ -3,13 +3,13 @@ FROM ${BASE_IMAGE}
 
 ARG MAVEN_VERSION=3.5.0
 ARG USER_HOME_DIR="/root"
-ARG SHA=beb91419245395bd69a4a6edad5ca3ec1a8b64e41457672dc687c173a495f034
+ARG SHA=d5a520ca8765ddbc86dca71249c602e2f798dedcc7430bc4979dd01918464c8dc69b694ec0dbbeeff6044179e1b98fce72af952663dd49503203d9742e328f3b
 ARG BASE_URL=https://archive.apache.org/dist/maven/maven-3/${MAVEN_VERSION}/binaries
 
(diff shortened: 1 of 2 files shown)


etsrepo:
I will all add java 11 support if https://github.com/GoogleCloudPlatform/cloud-builders/pull/437 is accepted

etsrepo:
fixes https://github.com/GoogleCloudPlatform/cloud-builders/issues/412 if this and https://github.com/GoogleCloudPlatform/cloud-builders/pull/437 merged

ImJasonH:
cc @loosebazooka

To avoid the need for these updates long-term, what do we think about making the image `FROM maven:$VERSION` and recommending people use the [official maven image](https://hub.docker.com/_/maven) directly?

```
steps
- name: maven:whatever-version-you-want
  args: ['--version']
```

This has been done for npm and yarn in #326 and I think it's been a positive change.

ImJasonH:
FWIW #451 updates the `mvn` builder to be `FROM maven:$VERSION` and recommends just using the `maven` image directly. Once that gets merged I'm not sure we need this PR, unless there's some reason you'd strongly prefer it.

ImJasonH:
@etsrepo `maven:latest` (which is currently `maven:3.6.0`) is already cached, so this PR should be unnecessary if you can just use the `maven` image directly.

Please let me know if this is incorrect or if you need anything else.

etsrepo:
@ImJasonH I agree on using official/direct images. thank you.

PR #451, opened 2019-02-04
Base mvn builder on official maven image

ImJasonH:
This reduces custom install logic for this builder and recommends users use the official `maven` image directly, which will remain up-to-date with future Maven releases and prevent the need for tedious and lapse-prone updates in this repo.

After this PR, `gcr.io/cloud-builders/mvn` will use whatever version of Maven and JDK is included in the official `maven:latest` image. Users who wish to pin to another supported version can use the official image of their choice. Tagged images for `3.3.6` and `3.5.0` are provided for convenience but will eventually stop being built.

This PR also removes caching of certain Maven packages, but since those cached packages were very old I doubt anybody really used them, and they required careful hand-maintenance to keep up-to-date (which we obviously didn't do).

This PR also updates yarn/README.md and npm/README.md to more forcefully/unformly recommend using an official image, and stops building `/java/` and `/nodejs/` which we've been meaning to do for a long time.

@coollog @aslo @loosebazooka


ImJasonH:
FYI I plan to make a similar change to the Gradle builder, if this LGTY'all

loosebazooka:
@chanseokoh

ImJasonH:
For context, Maven usage in GCB is very low, accounting for ~.12% of all executed build steps in the last 30d.

We actually already get more build steps that invoke the official Maven image tagged `maven:3.6-jdk-11`, so I suspect this would be low-impact, and with an easy migration path for whatever users might somehow be negatively affected.

mvn/README.md
```diff
@@ -1,10 +1,22 @@
 # Tool builder: `gcr.io/cloud-builders/mvn`
 
-This Cloud Build builder runs Maven. It also includes a number of dependencies
-that are precached within the image.
+This Cloud Build builder runs Maven.
+
+You should consider instead using an [official `maven`
+image](https://hub.docker.com/_/maven/) and specifying the `mvn` entrypoint:
```
bendory: yuck, at least in code review, this is rendering wrong due to the '_' in the path.

chanseokoh:
I just realized one thing missing from the official Maven images is the Docker, which was present in the old `cloud-builder/mvn` images. Jib could push images to the local Docker daemon and have GCB upload them to GCR easily, so it's a bit unfortunate that it's no longer doable. Not that I insist we should have Docker.

ImJasonH:
I'm a bit surprised that Jib support depends on having Docker installed, I thought the whole point was that Jib could push an image to the registry dockerlessly?

Either way, if you or anyone thinks that having Jib/Docker support in the builder is a requirement, we can back out this change and get it back.

chanseokoh:
@ImJasonH Jib doesn't depend on Docker. There was a user on GCB who wanted to publish an image directly to GCR but having an auth issue with GCR. And one workaround was to make Jib push to a local Docker daemon and have GCB upload that image to GCR. I'll add you to an email thread for details.

PR #452, opened 2019-02-04
Base gradle image on official gradle image

ImJasonH:
See also #451 for Maven

-----

**NB**: This PR currently does not build due to a permissions error, currently failing with:

```
starting container process caused "chdir to cwd (\"/workspace/examples/spring_boot\") set in config.json failed: permission denied": unknown.
```

I believe this is due to the fact that GCB executes containers as `root`. For some reason this isn't a problem with Maven.

Any help diagnosing this issue would be greatly appreciated.

-----

This reduces custom install logic for this builder and recommends users use the official `gradle` image directly, which will remain up-to-date with future Gradle releases and prevent the need for tedious and lapse-prone updates in this repo.

After this PR, gcr.io/cloud-builders/gradle will use whatever version of Gradle and JDK is included in the official `gradle:latest` image. Users who wish to pin to another supported version can use the official image of their choice. Tagged images for 4.6, 4.0 and 3.5 are provided for convenience but will eventually stop being built.

@coollog @aslo @loosebazooka @chanseokoh

### DIFF
--- a/gradle/Dockerfile
+++ b/gradle/Dockerfile
@@ -1,4 +1,5 @@
 ARG GRADLE_VERSION=latest
 FROM gradle:${GRADLE_VERSION}
+ENV GRADLE_USER_HOME "/root/.gradle/"
 ENTRYPOINT ["gradle"]
 
--- a/gradle/Dockerfile
+++ b/gradle/Dockerfile
@@ -1,26 +1,4 @@
-ARG BASE_IMAGE=gcr.io/cloud-builders/javac:8
-FROM ${BASE_IMAGE}
-
-ARG GRADLE_VERSION=4.0
-ARG USER_HOME_DIR="/root"
-ARG SHA=56bd2dde29ba2a93903c557da1745cafd72cdd8b6b0b83c05a40ed7896b79dfe
-ARG BASE_URL=https://services.gradle.org/distributions
-
-ENV GRADLE_HOME "/usr/share/gradle-${GRADLE_VERSION}"
-ENV GRADLE_USER_HOME "${USER_HOME_DIR}/.gradle/"
-
-RUN apt-get update -qqy && apt-get install -qqy curl \
-  && mkdir -p /usr/share "${GRADLE_USER_HOME}" \
-  && curl -fsSL -o "gradle-${GRADLE_VERSION}-bin.zip" "${BASE_URL}/gradle-${GRADLE_VERSION}-bin.zip" \
-  && echo "${SHA}  gradle-${GRADLE_VERSION}-bin.zip" | sha256sum -c - \
-  && unzip -qq "gradle-${GRADLE_VERSION}-bin.zip" \
-  && rm -f "gradle-${GRADLE_VERSION}-bin.zip" \
-  && mv "gradle-${GRADLE_VERSION}" /usr/share \
-  && ln -s "${GRADLE_HOME}/bin/gradle" /usr/bin/gradle \
-  && apt-get remove -qqy --purge curl \
-  && rm /var/lib/apt/lists/*_*
-
-ADD gradle.properties "${GRADLE_USER_HOME}"
-
-ENTRYPOINT ["/usr/bin/gradle"]
+ARG GRADLE_VERSION=latest
+FROM gradle:${GRADLE_VERSION}
+ENTRYPOINT ["gradle"]
 
--- a/gradle/Dockerfile
+++ b/gradle/Dockerfile
@@ -1,5 +1,6 @@
 ARG GRADLE_VERSION=latest
 FROM gradle:${GRADLE_VERSION}
 ENV GRADLE_USER_HOME "/root/.gradle/"
+USER root
 ENTRYPOINT ["gradle"]
 
--- a/gradle/cloudbuild.yaml
+++ b/gradle/cloudbuild.yaml
@@ -1,72 +1,52 @@
 # In this directory, run the following command to build this builder.
-# $ gcloud builds submit . --config=cloudbuild.yaml
-#
-# TODO(franklinn): Stop tagging java/mvn images once usage has dropped off.
+# $ gcloud builds submit
 
 steps:
+# Build old tagged images.
+# TODO(jasonhall): Deprecate and stop building these tagged images. Users
+# should use the official gradle image directly.
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
-  - '--build-arg=BASE_IMAGE=gcr.io/$PROJECT_ID/javac:8'
-  - '--build-arg=GRADLE_VERSION=4.6'
-  - '--build-arg=SHA=98bd5fd2b30e070517e03c51cbb32beee3e2ee1a84003a5a5d748996d4b1b915'
+  - '--build-arg=GRADLE_VERSION=4.6-jdk8'
   - '--tag=gcr.io/$PROJECT_ID/gradle:4.6-jdk-8'
-  - '--tag=gcr.io/$PROJECT_ID/java/gradle:4.6-jdk-8'
   - '.'
-  waitFor: ['-']
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
-  - '--build-arg=BASE_IMAGE=gcr.io/$PROJECT_ID/javac:8'
-  - '--build-arg=GRADLE_VERSION=4.0'
-  - '--build-arg=SHA=56bd2dde29ba2a93903c557da1745cafd72cdd8b6b0b83c05a40ed7896b79dfe'
+  - '--build-arg=GRADLE_VERSION=4.0-jdk8'
   - '--tag=gcr.io/$PROJECT_ID/gradle:4.0-jdk-8'
-  - '--tag=gcr.io/$PROJECT_ID/java/gradle:4.0-jdk-8'
   - '.'
-  waitFor: ['-']
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
-  - '--build-arg=BASE_IMAGE=gcr.io/$PROJECT_ID/javac:8'
-  - '--build-arg=GRADLE_VERSION=3.5'
-  - '--build-arg=SHA=0b7450798c190ff76b9f9a3d02e18b33d94553f708ebc08ebe09bdf99111d110'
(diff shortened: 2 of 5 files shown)


gradle/Dockerfile
```diff
@@ -1,26 +1,5 @@
...
-  && mv "gradle-${GRADLE_VERSION}" /usr/share \
-  && ln -s "${GRADLE_HOME}/bin/gradle" /usr/bin/gradle \
-  && apt-get remove -qqy --purge curl \
-  && rm /var/lib/apt/lists/*_*
-
-ADD gradle.properties "${GRADLE_USER_HOME}"
-
-ENTRYPOINT ["/usr/bin/gradle"]
+ARG GRADLE_VERSION=latest
+FROM gradle:${GRADLE_VERSION}
```
loosebazooka: I can only get this to work if I add
```
USER root
```

otherwise I get
```
Step #4: Failed to load native library 'libnative-platform.so' for Linux amd64.     
```

but I'm not really sure what the implications of this are.

loosebazooka: some context here: https://github.com/keeganwitt/docker-gradle/issues/29

loosebazooka:
and as a followup, the springboot example included in this repo is incompatible with gradle 5.2

ImJasonH:
> and as a followup, the springboot example included in this repo is incompatible with gradle `5.2`
> 
> EDIT: Also it looks like a `5.2` project by default will not be compatible with versions < `4.6`.

Well that's just great! I guess we could create a new example that is 5.2-compatible, and build that instead?

I've updated the Dockerfile to add `USER root` and commented out failing steps with details about why they're broken.

If a user using `gcr.io/cloud-builders/gradle` is using it to build a non-5.2-compatible codebase, it will fail after this release, and they'll have to use a tagged builder with an older version. Are we okay with that? Very very few users use the gradle builder today, even less than Maven, which is also not widely used.

loosebazooka:
Just for historical information:

The prime culprit appears to be setting 
```
ENV GRADLE_USER_HOME "/root/.gradle/"
```
without overriding the `USER` param in the base gradle image which is set to `gradle`.

loosebazooka:
>If a user using `gcr.io/cloud-builders/gradle` is using it to build a non-5.2-compatible codebase, it will fail after this release, and they'll have to use a tagged builder with an older version. Are we okay with that? Very very few users use the gradle builder today, even less than Maven, which is also not widely used.

I think this is your team's decision to make (but staying on latest by default should be an okay expectation).

LGTM

etsrepo:
@loosebazooka @ImJasonH what about java 11 support in this, using gradle with jdk:11 directly not working. so we need to create images for each version again?

etsrepo:
https://github.com/GoogleCloudPlatform/cloud-builders/pull/437

loosebazooka:
@etsrepo Gradle publishes Java 11 images. It is recommended you use those in your cloud build.yaml. These images are around for legacy reasons.

https://hub.docker.com/_/gradle

ImJasonH:
We need to resolve the `USER root` issue before Gradle images can be used unmodified in GCB

loosebazooka:
Oh I didn't realize that was a problem with the official images too?

ImJasonH:
Official images [set `USER gradle`](https://github.com/keeganwitt/docker-gradle/blob/9ecbf1925ddb905c0bf5993f10b173873c0432d6/jdk8/Dockerfile#L32), so they seem to be incompatible with GCB as-is.

I've updated this PR to note this incompatibility and we can hopefully address that in a future change.

chanseokoh:
I was wondering why it requires running as `root`, and it turns out Gradle running as the `gradle` user can't create the `build/` directory under `/workspace` owned by `root`. The following example demonstrates that we can use the official Gradle image as-is if it can create the `build/` directory.

```yaml
- name: 'bash'
  args: ['chown', '-R', '1000.1000', '/workspace/examples/spring_boot']
  dir: 'examples/spring_boot'
- name: 'gradle:4.6-jdk8'
  args: ['gradle', 'build']
  dir: 'examples/spring_boot'
```

So, it is not that there is an issue to be fixed on the side of the Gradle image. The doc says to run it as `root` if required:

> How to use this image
> Note that if you are mounting a volume and the uid running Docker is not `1000`, you should run as user `root` (`-u root`).

gradle/cloudbuild.yaml
```diff
@@ -1,79 +1,61 @@
...
+
+# TODO(jasonhall): The spring boot example is not compatible with gradle 5.2
+#- name: 'gcr.io/$PROJECT_ID/gradle'
+#  args: ['build']
+#  dir: 'examples/spring_boot'
+
+# Demonstrate using the gradle image directly.
+# TODO(jasonhall): This doesn't work because it doesn't have USER root
+# See https://github.com/keeganwitt/docker-gradle/issues/29
+#- name: 'gradle:4.6-jdk11-slim'
```
chanseokoh: Looks like there's no such tag.

gradle/Dockerfile
```diff
@@ -1,26 +1,6 @@
...
+ENV GRADLE_USER_HOME "/root/.gradle/"
```
chanseokoh: This doesn't seem necessary, unless there was a specific reason to have.

gradle/examples/spring_boot/Dockerfile
```diff
@@ -1,3 +0,0 @@
-FROM openjdk:8-jdk
```
chanseokoh: Since you removed this file, I think it makes sense to remove the corresponding entry in `examples/spring_boot/cloudbuild.yaml` too.
```yaml
steps:
- name: 'gcr.io/cloud-builders/gradle'
  args: ['build']
  args: ['build', '--tag=gcr.io/$PROJECT_ID/spring-boot', '.']
images: ['gcr.io/$PROJECT_ID/spring-boot']
```

Strat1987:
What's the latest status on supporting a newer gradle version using cloud build?

Mistic92:
Today I found that gradle shadowJar plugin 5.0 is not working with gradle 4 which cloud builder latest is based on and I have problem with running gradle image from docker hub. Is there any update with this?

gavingolden:
@Strat1987 @Mistic92 I finally got my build working by using the [gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) (`gradlew`) and a plain java image instead of a dedicated gradle image or the gcloud gradle builder.

I would rather not download gradle for each build but I had no luck with the suggestions above, so I'll use this for the time being. Is there a way to use the gradle wrapper and cache the gradle binary?

```yml
- name: 'java:8'
    entrypoint: "bash"
    args: [ "-c", "./gradlew build"]
- name: 'gcr.io/cloud-builders/gcloud'
     ...
```

This doesn't suffer from the file permissions issue that the gradle image has.

etsrepo:
if this is going no where in, why can't we merge pull request which supports java 8, 11 and gradle 3.x, 4.x, 5.x ? 
https://github.com/GoogleCloudPlatform/cloud-builders/pull/437

ippersolar:
Guys, on the Cloud Function kotlin sample repos, it is mentioning the usage of ShadowJar 5, which is not compatible with Gradle 4.6 anymore. Could you please bump up Gradle version ASAP for the cloud builder? You're out of sync now with GCP samples.

dsaff:
It appears that this issue was fixed with https://github.com/GoogleCloudPlatform/cloud-builders/commit/6913d60a006def3b046aa1356e9c02b665a114e3

But this doesn't seem to be reflected when running the public build step, yet.  Is there a release coming?

ImJasonH:
@chanseokoh That image has not been released yet, I expect it to be some time next week.

There's some confusion because GCB workers run a cached/released image even if the tag is `:latest`. Running this cloudbuild.yaml:

```yaml
steps:
- name: gcr.io/cloud-builders/gradle:latest
  args: ['--version']

```

```
Gradle 4.6
```

If you specify `:5.6.2-jdk-8` exactly, that image will be pulled and run:

```yaml
steps:
- name: gcr.io/cloud-builders/gradle:5.6.2-jdk-8
  args: ['--version']
```

```
Welcome to Gradle 5.6.2!
```

dsaff:
@chanseokoh ah, I see.  Yes, @ImJasonH also nails it--I was using latest and expecting gradle 5.6.2, but using the 5.6.2-jdk-8 tag directly does work.  Thanks.

chanseokoh:
@ImJasonH you said Gradle 5.6.2 would become available this week, but I see that GCB still runs on the cached Gradle 4.6 image. When will the cache be refreshed?

rafikk:
@chanseokoh, the production release is currently in progress. Gradle 5.6.2 should be available by week's end.

chanseokoh:
@rafikk 2 days ago, I occasionally hit 5.6.2 when using `gcr.io/cloud-builders/gradle`, and I was thinking the cached 4.6 image is being purged. But today, I tried more than 10 times, and it's always 4.6. Seems like a regression, and I don't really think 5.6.2 will be available by the end of the week. Any issue?

rafikk:
@chanseokoh, you're right. We had to roll back the cache update for an unrelated issue, and are now up against a release freeze for US holidays. Unfortunately, we won't be able to update the cached images until December. Sorry for the inconvenience, but for now you'll need to refer to the explicit version tag to use 5.6.2: gcr.io/cloud-builders/gradle:5.6.2-jdk-8.

chanseokoh:
@rafikk the holiday season is over. Any timeline for upgrading Gradle from 4.6 to 5.6.2?

ichaelm:
@chanseokoh To answer your question for @rafikk, the upgrade is finished, and I confirmed this by running a build with the image "gcr.io/cloud-builders/gradle".

chanseokoh:
Thanks @ichaelm!

And looks like this PR can move forward now (after fixing things). Apparently, recent official Gradle images now [run as `root`](https://github.com/GoogleCloudPlatform/cloud-builders/pull/611#issuecomment-586356585), so https://github.com/GoogleCloudPlatform/cloud-builders/pull/452#issuecomment-462932858 (and [this comment](https://github.com/GoogleCloudPlatform/cloud-builders/pull/452/files#diff-5f83fe68243f99b18b871a104e9c86faR7)) are no longer an issue.

bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/665

PR #453, opened 2019-02-04
Extend retries/backoff when fetching first file (the manifest)

squee1945:
When new projects are created, there is an eventual consistency latency that can cause first deployments to fail. This PR is a mitigation to retry longer for the manifest before failing.

@ImJasonH  @cynocracy


codrienne:
I ran the cloudbuild.yaml and it passes

PR #454, opened 2019-02-05
Base gcloud builder on official google/cloud-sdk image

ImJasonH:
See also #451 and #452 

This reduces custom install logic for this builder and recommends users use the official `google/cloud-sdk` image directly, which will remain up-to-date with future releases and prevent the need for tedious and lapse-prone updates in this repo.

This PR also updates README.md to more forcefully recommend using an official image, which enables version selection.

I've tested this with `gcloud`, `gsutil`, `git` and `kubectl`.


gcloud/README.md
```diff
@@ -1,18 +1,25 @@
...
 credentials of the [builder service
 account](https://cloud.google.com/cloud-build/docs/permissions) for the
-project.
+project. The latest released version of `gcloud` is used.
+
+You should consider instead using an [official `google/cloud-sdk`
+image](https://hub.docker.com/r/google/cloud-sdk) and specifying the `gcloud` entrypoint:
+
+```yaml
+steps:
```
bendory: Consider adding an example .yaml that explains "these two build steps are functionally equivalent" to make the difference between our image (with entrypoint) and the FROM image (without) explicit.

PR #455, opened 2019-02-05
Allow permission error to bubble up allowing cleaner output.

squee1945:
This keeps the permissionError from being wrapped and concealed from processing higher up the stack.

@codrienne PTAL


codrienne:
Change LGTM. For good measure, I'm running the gcs-fetcher/cloudbuild.yaml before merging.

Issue #290, opened 2018-06-07
kubectl image ignores env.CLOUDSDK_CONTAINER_CLUSTER when used in multiple steps


yungchin:
This may be one of those "don't do that, then" bugs, but just in case it's helpful: if you interact with multiple GKE clusters during the same build, like so

```
  - name: gcr.io/cloud-builders/kubectl
    args: ['create', '-f', 'this.yaml']
    env:
      - 'CLOUDSDK_COMPUTE_ZONE=europe-west1-c'
      - 'CLOUDSDK_CONTAINER_CLUSTER=my-first-cluster'
  - name: gcr.io/cloud-builders/kubectl
    args: ['create', '-f', 'that.yaml']
    env:
      - 'CLOUDSDK_COMPUTE_ZONE=europe-west1-c'
      - 'CLOUDSDK_CONTAINER_CLUSTER=my-other-cluster'
```

then I currently find that `that.yaml` is deployed to `my-first-cluster` instead of `my-other-cluster`.

ebrake:
I am also facing this issue: Updating CLOUDSDK_CONTAINER_CLUSTER does not get applied with multiple steps.

Philmod:
Hey!

I made a quick test, and the environment variables seem to work as expected: 
```
steps:
- name: gcr.io/cloud-builders/kubectl
  entrypoint: 'bash'
  args: ['-c', 'echo hello $$CLOUDSDK_CONTAINER_CLUSTER']
  env:
    - 'CLOUDSDK_CONTAINER_CLUSTER=my-first-cluster'
- name: gcr.io/cloud-builders/kubectl
  entrypoint: 'bash'
  args: ['-c', 'echo hello $$CLOUDSDK_CONTAINER_CLUSTER']
  env:
    - 'CLOUDSDK_CONTAINER_CLUSTER=my-other-cluster'
```

Logs: 
```
Starting Step #0
Step #0: Already have image (with digest): gcr.io/cloud-builders/kubectl
Step #0: hello my-first-cluster
Finished Step #0
Starting Step #1
Step #1: Already have image (with digest): gcr.io/cloud-builders/kubectl
Step #1: hello my-other-cluster
Finished Step #1
```

ebrake:
@Philmod Hm! Are you able to run commands against the cluster after the environment is updated? I am attempting to set an image for kubernetes in a separate cluster, but the initial CLOUDSDK_CONTAINER_CLUSTER is still being used.

[...]

PR #456, opened 2019-02-06
Update kubectl.bash to always get context

rajatvig:
When working with the `kubectl` builder across multiple clusters, the context is not switched as referenced in this issue - https://github.com/GoogleCloudPlatform/cloud-builders/issues/290.

The context should ideally be refreshed every run to use the variables as appropriate for that step.

### DIFF
--- a/kubectl/kubectl.bash
+++ b/kubectl/kubectl.bash
@@ -1,26 +1,24 @@
 #!/bin/bash
 
-# If there is no current context, get one.
-if [[ $(kubectl config current-context 2> /dev/null) == "" ]]; then
-    cluster=$(gcloud config get-value container/cluster 2> /dev/null)
-    zone=$(gcloud config get-value compute/zone 2> /dev/null)
-    project=$(gcloud config get-value core/project 2> /dev/null)
+# always get a new context
+cluster=$(gcloud config get-value container/cluster 2> /dev/null)
+zone=$(gcloud config get-value compute/zone 2> /dev/null)
+project=$(gcloud config get-value core/project 2> /dev/null)
 
-    function var_usage() {
-        cat <<EOF
+function var_usage() {
+    cat <<EOF
 No cluster is set. To set the cluster (and the zone where it is found), set the environment variables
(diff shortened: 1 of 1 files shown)


ImJasonH:
@DazWilkin WDYT? Or is there someone else on the GKE team we should ask?

DazWilkin:
LGTM

I'm curious to understand why the gate was added to `kubectl.bash` in the first place. It seems overly specific but perhaps was from an abundance of caution; I'm hoping there's no path that will now break (but think there isn't).

Personally, I dislike using the environment as a side-effect for state changes and prefer to see `--zone=${ZONE}` and `--region=${REGION}` for the specificity on commands.

BUT this would be challenging with `--context=${CONTEXT}` because it requires a lookup (or perhaps filtering?) to determine the correct context.

edvinerikson:
This change seem to break our CI builds with this error:
```
Step #1: No cluster is set. To set the cluster (and the zone where it is found), set the environment variables
Step #1: CLOUDSDK_COMPUTE_ZONE=
Step #1: CLOUDSDK_CONTAINER_CLUSTER=```
```

We use the tool like this, and I am wondering if this is not supported?

```
  - name: gcr.io/cloud-builders/gcloud
    env:
      - "KUBECONFIG=/kube/config"
    volumes:
      - name: "kube"
        path: /kube
    args:
      - "container"
      - "clusters"
      - "get-credentials"
      - "leo-frontend"
      - "--region=europe-west1-c"
      - "--project=leo-dev-shared"

  - name: gcr.io/cloud-builders/kubectl
    env:
      - "KUBECONFIG=/kube/config"
    volumes:
      - name: "kube"
        path: /kube
    args:
      - config
      - current-context
```

farrellit:
Also broke mine.  I think because  we're doing Regional clusters and yet this PR seems to assume that either the zone will be set, the zone ENV will be set or the script has to bomb out.

daanemanz:
Broke our builds as well. Similar to @edvinerikson we're setting the `KUBECONFIG` environment variable. See https://github.com/GoogleCloudPlatform/cloud-builders/issues/466.

PR #462, opened 2019-02-21
use official gcloud:alpine image for gcloud-slim

oliverxchen:
#### What this PR does:

Partially un-reverts @ImJasonH's revert in #459:

* bases `cloud-builders/gcloud-slim` on the official `google/cloud-sdk:alpine` docker image, which was updated a couple days ago to use Alpine 3.9. Alpine 3.9 can install git version 2.20.1, which avoids the problem causing the need for the original revert. 
* `cloud-sdk:latest` still only has git version 2.11, so `cloud-buiders/gcloud` still needs to be a custom install. The installation of JRE has been changed to work on Alpine.
* `cloud-builders/kubectl` was changed to use `google/cloud-sdk:alpine`.

#### Effect of PR:

Less maintenance will be needed for the gcloud-slim image, and there are also substantial reductions in image sizes, which doesn't matter much for running on Cloud Build, but is good when running in other environments. These image sizes were observed locally using `docker images`:

* gcloud-slim: 972MB -> 220MB
* gcloud: 2.51GB -> 1.4GB
* kubectl: 1.43GB -> 621MB

#### Validation

* Git version on gcloud and gcloud-slim:
   ```
    $ docker run -ti --entrypoint git gcloud --version
    git version 2.20.1
    $ docker run -ti --entrypoint git gcloud-slim --version
    git version 2.20.1
    ```
* I tried running the datastore emulator on the gcloud image to test the JRE install and that worked:
    ```
    $ docker run -ti --entrypoint /bin/sh gcloud
    / # gcloud config set project foo
    Updated property [core/project].
    / # gcloud beta emulators datastore start
   ...
   [datastore] Dev App Server is now running.
    ```
* I tried this new kubectl image in a cloud build step in my own project and it worked


oliverxchen:
Hmmm... thought about this a bit more and it would probably break a lot of images that use these three images as bases. How about keeping `cloud-builders/gcloud`, `cloud-builders/gcloud-slim` and `cloud-builders/kubectl` as is, and just having a new image `cloud-builders/gcloud-alpine`?

bendory:
closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #481, opened 2019-04-25
Add top-level cloudbuild.yaml

icco:
This adds a generic cloudbuild.yaml to the root directory so that if you run `gcloud builds submit` in the root folder, or have gcr auto builds for your repo, this will automatically start all of the builds. This is useful when developing or testing cloud-builders in your own repos.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,11 @@
+steps:
+- name: 'gcr.io/cloud-builders/gcloud'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    for config in $(ls */cloudbuild.yaml); do
+      d=$(dirname "${config}")
+      echo "=> $config ... "
+      gcloud builds submit $d --config=${config} --async --quiet
+    done


cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
+steps:
+- name: 'gcr.io/cloud-builders/gcloud'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    for config in $(ls */cloudbuild.yaml); do
+      d=$(dirname "${config}")
+      echo "=> $config ... "
+      gcloud builds submit $d --config=${config} --async --quiet
```
ImJasonH: `--async` here means the test won't actually wait for the builds to finish, so you won't be able to use it to determine whether they all succeed. Some might not ever even start executing, due to concurrency caps.

Of course, without `--async` they'll run in serial and this build will take a very long time. 😢 

I could imagine some terrible hack where we tag each of these builds with some unique nonce, run them with `--async` then poll `gcloud builds list --filter=tag=${nonce}` and see that they all succeed. 

Or, to crib from [what's done in the community repo](https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/cloudbuild.yaml), start each build without `--async` in a background process and [`wait`](https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/cloudbuild.yaml#L45) until they all complete.

WDYT?

icco: I personally don't think there's a need to wait, given the status is in all of the other builds, and the fact that some of the builds in this repo can never be built. I first actually implemented it similarly (using https://stackoverflow.com/questions/51861870/github-cloud-build-integration-with-multiple-cloudbuild-yamls-in-monorepo as a base) but it seemed pretty useless given how unreadable the logs were and how many of the builds were broken.

icco:
ping @ImJasonH

cloudbuild.yaml
ImJasonH: Sorry for the lack of response. I'm not sure what the value of this mega-build is if we don't wait to see whether builds complete successfully.

Which builds were broken? We regularly run every build in this repo as part of an automated release, and we very rarely see build breakages for these builders.

Unreadable logs I can definitely understand. Maybe we could start the builds with `--async`, wait for all of them to complete, then `gcloud bulids log` each (falied?) build separately so that they're not interleaved?

bendory:
In the context of the [deprecation and upcoming deletion of the vast majority (all?) of these builders](https://github.com/GoogleCloudPlatform/cloud-builders/issues?q=is%3Aissue+label%3Adeprecation), this is no longer relevant.

Closing as obsolete.

PR #482, opened 2019-04-30
Cleanup all files from /tmp

fvoznika:
Remove files that may have been created under /tmp
during image build.

```bash
# find /tmp
/tmp
/tmp/.X11-unix
/tmp/hsperfdata_root
/tmp/.Test-unix
/tmp/.XIM-unix
/tmp/build_event_text_file.RdWTd
/tmp/.font-unix
/tmp/.ICE-unix
```

### DIFF
--- a/bazel/Dockerfile
+++ b/bazel/Dockerfile
@@ -52,4 +52,7 @@ RUN \
 RUN mkdir -p /workspace
 RUN echo 'startup --output_base=/workspace/.bazel' > ~/.bazelrc
 
+# Cleanup
+RUN rm -rf /tmp && mkdir /tmp && chmod a+rwxt /tmp
+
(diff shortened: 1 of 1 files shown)


ImJasonH:
Can you describe what bug this fixes? Invoking `bazel` normally (outside of a container or GCB) doesn't clear out `/tmp` so I'm not sure why this packaging of the tool should do it.

fvoznika:
If you run this container with gVisor, for example, the fact that /tmp is empty allows the sandbox to mount it as an internal tmpfs to make it faster. Apart from that, it's good practice to keep only files strictly required inside the image.

bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/667

PR #489, opened 2019-05-10
Base bazel builder on official bazel image

ImJasonH:
cc @nlopezgi

This removes custom installation logic and ensures we remain based on the official Bazel installation instructions and image release schedule.

We still need Git >2.0.1 and Docker installed, which we'll install manually ourselves for now, as well as the wrapper script to expose invocation IDs via build step outputs.


ImJasonH:
Closing in favor of #504

PR #490, opened 2019-05-14
Update tagged versions in readme

orinokai:
It looks like the tagged versions at gcr.io/cloud-builders/docker differ from those mentioned in the readme. The readme is older than the new Dockerfile in the repo so I'm guessing it needs an update?


bendory:
thanks!

PR #494, opened 2019-05-20
Upgrade Docker to 18.09.6.

bendory:


codrienne:
I think the Docker version in cloud-builders/javac/cloudbuild.yaml also needs to be changed.

PR #500, opened 2019-06-10
Allow usage of private golang repositories

mlesar:
Added `GIT_CONFIG_VAL` environment variable with which you can set for example GitHub access token so that you can get the code from private repositories without using `vendor` dir.
Example:
```
  env:
    - 'GO111MODULE=on'
    - 'GIT_CONFIG_VAL=url."https://${_GITHUB_ACCESS_TOKEN}@github.com/".insteadOf "https://github.com/"'
```


mlesar:
I signed it!

PR #504, opened 2019-06-20
Remove special-case logic from Bazel builder

ImJasonH:
Instead, base it on the official Bazel image. Recommend that users use the official image directly, which enables version selection.


bendory:
Rather than merging this, I'm leaving the existing `gcr.io/cloud-builders/bazel` in place as the final release of this image. Per https://github.com/GoogleCloudPlatform/cloud-builders/issues/667, the deprecation points users to the replacement.

Any users who rely on the special-case logic will be able to continue to build with the deprecated image (which will age out over time) or can build a special-case builder for their custom usage with the replacement bazel image as their base image.

PR #510, opened 2019-07-08
Add `gke-deploy` builder.

joonlim:


joonlim:
@ImJasonH

gke-deploy/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:stretch AS build-env
+RUN go get github.com/gobuffalo/packr/v2/packr2
+ADD . /go/src/github.com/GoogleCloudPlatform/cloud-builders/gke-deploy
+RUN go get -d github.com/GoogleCloudPlatform/cloud-builders/gke-deploy
+RUN cd /go/src/github.com/GoogleCloudPlatform/cloud-builders/gke-deploy && packr2
```
ImJasonH: What does `packr2` do?

gke-deploy/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN go build -o /gke-deploy github.com/GoogleCloudPlatform/cloud-builders/gke-deploy
+
+FROM google/cloud-sdk:alpine
+RUN gcloud -q components install kubectl
```
ImJasonH: Does this tool need to shell out to `kubectl`, or can it make direct API calls to the cluster, to remove this dependency?

gke-deploy/README.md
```diff
@@ -0,0 +1,45 @@
...
+following:
+
+### Prepare Step
+
+1. `gke-deploy` modifies a set of Kubernetes resource YAML configs to use a
+   container image's digest instead of a tag and adds [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications).
+
+### Apply Step
+
+2. `gke-deploy` gets authorized to access a GKE cluster.
```
ImJasonH: Is it intentional that this numbering continues from the above section?

gke-deploy/cloudbuild.yaml
```diff
@@ -0,0 +1,16 @@
...
+# Build the gke-deploy binary and put into the builder image.
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/gke-deploy', '.']
+- name: 'gcr.io/$PROJECT_ID/gke-deploy'
+  args: ['--help']
+
+images:
+- 'gcr.io/$PROJECT_ID/gke-deploy'
+
+timeout: '2100s'
```
ImJasonH: Is this needed? How long does this normally take?

gke-deploy/Dockerfile
joonlim: `packr2` is used to embed static files as go strings. I use it to create in-memory kubernetes objects from files at startup.

gke-deploy/Dockerfile
joonlim: Yes, we decided to using `kubectl` was the best solution for us to get this out the door.

gke-deploy/README.md
joonlim: Fixed.

gke-deploy/cloudbuild.yaml
joonlim: This was a remnant from when I had a functional test here before. I have now removed it.

gke-deploy/Dockerfile
ImJasonH: Hmm, this seems like a fairly large dependency to take on just to embed [these five lines](https://github.com/GoogleCloudPlatform/cloud-builders/blob/c26aeb408bd15b51e9a808958f16988946e8845c/gke-deploy/core/resource/templates/namespace.yaml)?

Unless you plan on embedding a few dozen more templates -- and even then until you actually do -- you might want to just embed the string directly, in Go code:

```go
const namespaceTemplate = `
apiVersion: v1
kind: Namespace
metadata:
  name:
`
```

If you do decide to keep packr2 as a dependency, you might want to vendor it so you can more closely control the version of packr2 that's used to embed the Go code.

cc @stewchen in case this came up during the original code review.

ImJasonH: Just to check whether it would work, I inlined namespace.yaml and removed the packr library dependency, which removed 258k lines of dependencies!

ImJasonH: In the interest of getting this in as a simple copy from the other repo, we can just merge this as-is and remove the dependency as a followup PR. Don't worry about trying to make this change in this PR.

joonlim: That's fair. I do have plans to add at least one more template, but nowhere near a dozen. I'll create a ticket for this refactoring.

PR #512, opened 2019-07-09
Fix parsing of image name by using go-containerregistry library.

joonlim:


joonlim:
@ImJasonH @stewchen

gke-deploy/deployer/deployer.go
```diff
@@ -53,11 +53,15 @@ func (d *Deployer) Prepare(ctx context.Context, images []string, appName, appVer
 	fmt.Printf("Configs to prepare: %v\n", objs)
 
 	for _, image := range images {
+		// name.Digest
```
ImJasonH: This looks like leftover debugging?

gke-deploy/core/container/container.go
```diff
@@ -17,17 +17,83 @@ import (
...
 
 const (
 	imageDigestFormat = "value(image_summary.digest)"
 )
 
-// GetDigest uses an image string to get its corresponding digest.
-func GetDigest(ctx context.Context, image string, gs services.GcloudService) (string, error) {
-	digest, err := gs.ContainerImagesDescribe(ctx, image, imageDigestFormat)
+// Image is an alias to name.Reference.
+type Image = name.Reference
```
ImJasonH: Is this alias useful? Can you just pass around `name.Reference`s wherever needed instead?

(Technically, "container" is a misnomer here anyway, you're dealing with "images", not any running "containers")

gke-deploy/core/container/container.go
```diff
@@ -17,17 +17,83 @@ import (
...
+			return nil, fmt.Errorf("duplicate image name: %q", imName)
+		}
+		exists[imName] = true
+		ims = append(ims, im)
+	}
+
+	return ims, nil
+}
+
+func parseImage(ctx context.Context, image string) (Image, error) {
```
ImJasonH: `ctx` is unused in this method, if you drop it you can also drop it from `ParseImages` above.

gke-deploy/core/container/container.go
```diff
@@ -17,17 +17,83 @@ import (
...
+	im, err := parseImage(ctx, image)
 	if err != nil {
-		return "", fmt.Errorf("failed to get image digest: %v", err)
+		return "", err
+	}
+	return GetName(ctx, im)
+}
+
+// GetName gets an image's name.
+func GetName(ctx context.Context, image Image) (string, error) {
```
ImJasonH: `ctx` is unused in this method, if you drop it you can also drop it from `GetNameFromString` above.

gke-deploy/core/container/container.go
```diff
@@ -17,17 +17,83 @@ import (
...
+func parseImage(ctx context.Context, image string) (Image, error) {
+	im, err := name.ParseReference(image)
+	if err != nil {
+		return nil, fmt.Errorf("image is invalid: %q", image)
+	}
+	return im, nil
+}
+
+// GetNameFromString gets an image's name, given a string representation of the image.
+func GetNameFromString(ctx context.Context, image string) (string, error) {
```
ImJasonH: This name could use some work. How about "ParseImageReference"? Something that indicates the input is a string, representing an image reference, and it returns (I think?) the fully-qualified image reference after normalizing it?

gke-deploy/core/container/container.go
ImJasonH: I think `GetName` could just be `ref.String()` and remove this method.

gke-deploy/core/container/container.go
```diff
@@ -17,17 +17,83 @@ import (
...
+	default:
+		return "", fmt.Errorf("invalid image type: %s", t)
+	}
+}
+
+// GetDigest gets an image's corresponding digest.
+func GetDigest(ctx context.Context, image Image, gs services.GcloudService) (string, error) {
+	switch t := image.(type) {
+	case name.Tag:
+		digest, err := gs.ContainerImagesDescribe(ctx, t.Name(), imageDigestFormat)
```
ImJasonH: There are better ways to get this information than shelling out to `gcloud container images describe` -- like for instance, go-containerregistry's `v1/remote.Image()` method which returns a `v1.Image` with a `Digest` method.

I'm considering adding a `remote.Resolve(name.Tag) (name.Digest, error)` method to streamline this since it's a pretty common workflow.

gke-deploy/core/container/container_test.go
```diff
@@ -15,35 +15,215 @@ package container
...
+	image2 := "gcr.io/my-project/my-image-2@sha256:929665b8eb2bb286535d29cd73c71808d7e1ad830046333f6cf0ce497996eb79"
+	image3 := "gcr.io/my-project/my-image-3"
+
+	tests := []struct {
+		name string
+
+		images []string
+
+		want []Image
+	}{
```
ImJasonH: nit: To preserve vertical and horizontal space, you can put these `{`s on the same line, e.g.:

```go
tests := []struct{
  ...struct definition...
}{{
  ...case...
}, {
  ...case...
}, {
  ...case...
}}
```

gke-deploy/core/resource/testing/deployment-3.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: extensions/v1beta1
```
ImJasonH: Are these YAMLs included in this PR on purpose?

gke-deploy/core/container/container.go
joonlim: Removed the reference.

gke-deploy/core/container/container.go
joonlim: I added a comment indicating what it does.
"given "gcr.io/my-project/my-app:1.0.0", returns "gcr.io/project/my-app""

joonlim: Also renamed to "ParseImageReference"

gke-deploy/core/container/container.go
joonlim: This method returns the image string portion from the start to the tag/digest.
e.g., given "gcr.io/my-project/my-app:1.0.0", returns "gcr.io/project/my-app

I couldn't find an interface method that does this for me so I implemented it like so.
ref.String() returns the entire image string, including the tag/digest

gke-deploy/core/container/container.go
joonlim: Thanks for the advice. I have implemented your suggestion.

gke-deploy/core/container/container_test.go
joonlim: I agree that this will be better. If it's okay with you, I'd like to do this in a future PR in order to touch all existing tests and not pollute this PR.

gke-deploy/core/resource/testing/deployment-3.yaml
joonlim: These YAMLs are used in unit tests

gke-deploy/core/image/image.go
```diff
@@ -47,31 +40,32 @@ func ParseImages(ctx context.Context, images []string) ([]Image, error) {
...
-		ims = append(ims, im)
+		refs = append(refs, ref)
 	}
 
-	return ims, nil
+	return refs, nil
 }
 
-func parseImage(ctx context.Context, image string) (Image, error) {
+func parseImage(image string) (name.Reference, error) {
```
ImJasonH: This method doesn't really add much. It adds a small prefix to the error string, but I'd argue that's not enough to justify having this method. Just call `name.ParseReference` directly in `ParseImages` and remove this method.

gke-deploy/core/image/image.go
```diff
@@ -47,31 +40,32 @@ func ParseImages(ctx context.Context, images []string) ([]Image, error) {
...
 	}
 	return im, nil
 }
 
-// GetNameFromString gets an image's name, given a string representation of the image.
-func GetNameFromString(ctx context.Context, image string) (string, error) {
-	im, err := parseImage(ctx, image)
+// ParseImageReference gets an image's name, given a string representation of the image.
+// e.g., given "gcr.io/my-project/my-app:1.0.0", returns "gcr.io/project/my-app"
+func ParseImageReference(image string) (string, error) {
```
ImJasonH: `ParseImageReference` isn't very descriptive here either. It parses, but then it also returns only part of what it parsed out.

gke-deploy/core/image/image.go
```diff
@@ -47,31 +40,32 @@ func ParseImages(ctx context.Context, images []string) ([]Image, error) {
...
 		return "", err
 	}
-	return GetName(ctx, im)
+	return GetName(im)
 }
 
 // GetName gets an image's name.
-func GetName(ctx context.Context, image Image) (string, error) {
+func GetName(image name.Reference) (string, error) {
 	switch t := image.(type) {
```
ImJasonH: `name.Reference` has all the methods you need, and you don't actually need to care whether it's a `name.Tag` or a `name.Digest` (or a future type that doesn't exist yet). Just `return fmt.Sprintf("%s/%s", ref.RegistryStr(), ref.RepositoryStr())`, and at this point `GetName` doesn't even need to return an error.

gke-deploy/core/image/image.go
```diff
@@ -83,17 +77,14 @@ func GetName(ctx context.Context, image Image) (string, error) {
...
-		digest, err := gs.ContainerImagesDescribe(ctx, t.Name(), imageDigestFormat)
-		if err != nil {
-			return "", fmt.Errorf("failed to get image digest: %v", err)
-		}
-		return digest, nil
-	case name.Digest:
-		return image.Identifier(), nil
-	default:
-		return "", fmt.Errorf("invalid image type: %s", t)
+func GetDigest(ctx context.Context, image name.Reference, rs services.RemoteService) (string, error) {
```
ImJasonH: I'd suggest something like `ResolveReference`, since this returns a fully-qualified string containing the image name and digest, not just the digest.

gke-deploy/core/image/image.go
```diff
@@ -47,31 +40,32 @@ func ParseImages(ctx context.Context, images []string) ([]Image, error) {
 			return nil, fmt.Errorf("duplicate image name: %q", imName)
```
ImJasonH: off-topic: Are duplicates that big of a problem here? I don't know how this method is called, but without any more context I would have guessed this would just strip duplicates, rather than returning an error.

If this doesn't have to return an error, and `GetName` doesn't return any error, `ParseImages` can also not return an error.

Worth considering for a future PR.

gke-deploy/core/image/image.go
joonlim: https://github.com/google/go-containerregistry/blob/master/pkg/name/ref.go#L24
name.Reference doesn't have RegistryStr and RepositoryStr methods.

gke-deploy/core/image/image.go
joonlim: This only returns the digest

joonlim: But, looking at my code, it can return the fully-qualified string rather than just the digest. I will change this.

gke-deploy/core/image/image.go
ImJasonH: Sorry, you're right, `ref.Context().RegistryStr()` and `ref.Context().RepositoryStr()`

https://play.golang.org/p/Da8RxUZTlvL

gke-deploy/core/image/image.go
joonlim: It is more of a nice-to-have. Allowing duplicates means the order of the image flag matters.
e.g., `-i image:1 -i image:2` will lead to containers pointing to image:2 and  `-i image:2 -i image:1` will lead to containers pointing to image:1

I can also opt into checking for duplicates at a higher layer closer to main.

gke-deploy/core/image/image.go
joonlim: If name.Reference has RegistryStr and RepositoryStr methods I can remove the GetName method

joonlim: ^ Sweet! I'll get to refactoring

PR #518, opened 2019-07-10
[Go] Modules sharing packages example

DazWilkin:
@ImJasonH  -- added an example.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,5 @@
+module github.com/DazWilkin/cloud-builders
+
+go 1.12
+
+require github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/DazWilkin/cloud-builders
```
ImJasonH: Is this file necessary? It should be enough to have the go.mod in `go/examples/multi_step/` right?

DazWilkin: Hmmm... user error.

DazWilkin: Apologies. Fixed: https://github.com/GoogleCloudPlatform/cloud-builders/pull/518/commits/ecb516fc9775eed1d5c9926c4a71054765569c3e

PR #520, opened 2019-07-15
Clean up gke-deploy code and docs.

joonlim:


joonlim:
@stewchen Can you review please?

PR #521, opened 2019-07-15
Mitigate race condition between goroutines when downloading files with same SHA hash.

squee1945:


gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -198,6 +199,12 @@ func (gf *Fetcher) fetchObject(ctx context.Context, j job) *jobReport {
 	var tmpfile string
 	var backoff time.Duration
 
+	// Within a manifest, multiple files may have the same SHA. This can lead
+	// to a race condition within the goworkers that are downloading the files
+	// concurrently. To mitigate this issue, we add some randomness to the name
```
ImJasonH: Does this mean the file will be downloaded multiple times, instead of being downloaded once and copied to multiple locations?

squee1945: Yes. I don't think it's worth tearing apart the structure of this code to solve this particular edge case.

PR #524, opened 2019-07-17
Add expose flag, which creates a Service resource.

joonlim:


joonlim:
@stewchen Please review

gke-deploy/cmd/run/run.go
```diff
@@ -86,6 +87,7 @@ func NewRunCommand() *cobra.Command {
 	cmd.Flags().StringSliceVarP(&options.labels, "label", "L", nil, "Label(s) to add to Kubernetes resources (k1=v1). Labels can be set comma-delimited or as separate flags. If two or more labels with the same key are listed, the last one is used.")
 	cmd.Flags().StringVarP(&options.namespace, "namespace", "n", "default", "Name of GKE cluster to deploy to.")
 	cmd.Flags().StringVarP(&options.output, "output", "o", "./output", "Target directory to store modified Kubernetes resource configs.")
+	cmd.Flags().IntVarP(&options.exposePort, "expose", "x", 0, "Creates a Service resource that connects to a deployed resource using a selector that matches the value provided by --app. The port provided will be used to expose the deployed resource (i.e., port and targetPort will be set to the value provided in this flag).")
```
stewchen: call out the label selector key here as well

gke-deploy/core/resource/templates.go
```diff
@@ -16,6 +16,20 @@ const (
...
 kind: Namespace
 metadata:
-  name:
+  name: %s
+`
+
+	serviceTemplate = `apiVersion: v1
+kind: Service
+metadata:
+  name: %s
```
stewchen: Do we need to specify the namespace of this service here? This should match the target cluster namespace

gke-deploy/deployer/deployer.go
```diff
@@ -47,7 +47,7 @@ type Deployer struct {
 }
 
 // Prepare handles preparing deployment.
-func (d *Deployer) Prepare(ctx context.Context, images []name.Reference, appName, appVersion, config, output, namespace string, labels map[string]string) error {
+func (d *Deployer) Prepare(ctx context.Context, images []name.Reference, appName, appVersion, config, output, namespace string, labels map[string]string, exposePort int) error {
```
ImJasonH: As a general comment, these args could be encapsulated into a struct, maybe even the `options` struct they originally come from, exported and moved into this package. Not urgent, just something to consider.

gke-deploy/doc/gke-deploy_prepare.md
```diff
@@ -32,6 +32,7 @@ gke-deploy prepare [flags]
 
 ```
   -a, --app string         Application name of the Kubernetes deployment.
+  -x, --expose int         Creates a Service resource that connects to a deployed resource using a selector that matches the value provided by --app. The port provided will be used to expose the deployed resource (i.e., port and targetPort will be set to the value provided in this flag).
   -f, --filename string    Config file or directory of config files to use to create the Kubernetes resources (file or files in directory must end with ".yml" or ".yaml").
   -h, --help               help for prepare
   -i, --image strings      Image(s) to be deployed. Images can be set comma-delimited or as separate flags.
```
stewchen: Unrelated to this PR but the comment for `-namespace` should be "Namespace of GKE cluster to deploy to"

gke-deploy/cmd/prepare/prepare.go
```diff
@@ -75,6 +76,7 @@ func NewPrepareCommand() *cobra.Command {
 	cmd.Flags().StringSliceVarP(&options.labels, "label", "L", nil, "Label(s) to add to Kubernetes resources (k1=v1). Labels can be set comma-delimited or as separate flags. If two or more labels with the same key are listed, the last one is used.")
 	cmd.Flags().StringVarP(&options.namespace, "namespace", "n", "default", "Name of GKE cluster to deploy to.")
 	cmd.Flags().StringVarP(&options.output, "output", "o", "./output", "Target directory to store modified Kubernetes resource configs.")
+	cmd.Flags().IntVarP(&options.exposePort, "expose", "x", 0, "Creates a Service resource that connects to a deployed resource using a selector that matches the value provided by --app. The port provided will be used to expose the deployed resource (i.e., port and targetPort will be set to the value provided in this flag).")
```
stewchen: call out the label selector key here?
e.g. "using a selector that matches the label with key as `app.kubernetes.io/name` and value provided by -app"

gke-deploy/deployer/deployer.go
joonlim: The parameter list is getting pretty long for this function. I'll consider doing that the next time a new arg is added.

gke-deploy/core/resource/templates.go
joonlim: Spoke in person

PR #525, opened 2019-07-17
Warn user if Service config already exists.

joonlim:


joonlim:
@stewchen please review

PR #527, opened 2019-07-19
Allow gke-deploy to work when gcloud is not in PATH.

joonlim:


joonlim:
@stewchen please review

Issue #478, opened 2019-04-15
kubectl builder does not work with regional clusters


edruela:
I'm working in a CI/CD pipeline using Cloud Build and Cloud Run.
Because of a current limitation on the Cloud Run, I need to create the namespace in order to deploy the application. 

I was trying to run a step to execute a kubectl create namespace, but got an issue while trying to do it. My K8S cluster is regional and builder is failing when I don't provide a value for CLOUDSDK_COMPUTE_ZONE. 

But if I don't provide the value the build fail with the next message:

No cluster is set. To set the cluster (and the zone where it is found), set the environment variables
CLOUDSDK_COMPUTE_ZONE=<cluster zone>
CLOUDSDK_CONTAINER_CLUSTER=<cluster name>

ENV variables set by the yaml file:
Env. Vars	
CLOUDSDK_COMPUTE_REGION=us-central1 CLOUDSDK_CONTAINER_CLUSTER=my-cluster

codrienne:
I believe this is Working As Intended, but maybe we to make the builder more flexible to permit other workflows. In the meantime, I would suggest building your own kubectl buildstep that is based on our official gcr.io/cloud-builders/kubectl. Our official builders are cached, so it will be very fast in pulling the layers.

Example: https://github.com/codrienne/kubectl-buildstep

edruela:
@codrienne,

I will fork the repository and create my own builder, but what I'm reporting is that the official builder don't work with Regional Cluster.

You can easily validate that, implementing a test providing only the CLOUDSDK_COMPUTE_REGION and CLOUDSDK_CONTAINER_CLUSTER as described in the documentation and you will see the builder will return an error.

Thanks,
@edruela

codrienne:
Gotcha, I've renamed the Github issue.

ptinsley:
+1 on this, I had to write my own step to work around this and was coming here to file an issue.

icco:
@codrienne we added support to the builder here in github, but it doesn't seem to exist in the versions that are in GCR. https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/kubectl/kubectl.bash has the code to make regional clusters work...

bendory:
closing as obsolete; context https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #530, opened 2019-07-23
kubectl: Always get GKE credentials

### LINKED ISSUE #465: kubectl no longer honors credentials from `gcloud container clusters get-credentials`
# Affected builder image

gcr.io/cloud-builders/kubectl

## Expected Behavior

Previous step running `gcloud container clusters get-credentials` used to expose credentials in such a way that they'd be picked up by a subsequent `kubectl` step

## Actual Behavior

Recently (not exactly sure when) the `kubectl` invocation started erroring out with:

`No cluster is set. To set the cluster (and the zone where it is found), set the environment variables`

1. have an existing kubernetes gke cluster
2. run the `glcoud container clusters get-credentials` from a first cloudbuild step for that cluster
3. run the 'kubectl' builder and you'll get the error

## Additional Info

This seems like a regression, but maybe I've just been doing it wrong and it happened to work, IDK.  Kubectl builder should imho be agnostic as to how it gets its credentials.  Setup of credentials should be optional.  Not sure why this stopped working.

ImJasonH:
Fixes #465 #290 #478 

Previous attempt in #456 was reverted in #467 because it didn't support regional clusters (#478), but was never rolled forward with a fix for regional clusters. This PR addresses this shortcoming, tested with my own project for both regional and zonal clusters, with multiple steps to check that context is overridden at each step.

### DIFF
--- a/kubectl/kubectl.bash
+++ b/kubectl/kubectl.bash
@@ -11,9 +11,6 @@ EOF
   exit 1
 }
 
-# If there is no kubectl context, use gcloud to get one, otherwise we can
-# assume the user hase configured the context with some kubectl related env
-# var.
 cluster=$(gcloud config get-value container/cluster 2> /dev/null)
 region=${CLOUDSDK_COMPUTE_REGION:-$(gcloud config get-value compute/region 2> /dev/null)}
 zone=$(gcloud config get-value compute/zone 2> /dev/null)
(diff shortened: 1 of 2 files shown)


kubectl/kubectl.bash
```diff
@@ -14,23 +14,21 @@ EOF
 # If there is no kubectl context, use gcloud to get one, otherwise we can
```
ichaelm: This comment is made obsolete.

PR #532, opened 2019-07-23
Use default keychain for creating remote Image reference.

joonlim:


joonlim:
@stewchen @ImJasonH Please review

PR #540, opened 2019-07-31
Fix gke-deploy deploying-with-cloud-build codelab.

joonlim:


joonlim:
@dorbin @ImJasonH Please review

PR #543, opened 2019-08-05
Support .tar.gz files in gcs-fetcher

ImJasonH:
* Support new `-type` flag value, `TarGzArchive` which ungzips and untars the archive into the destination directory.
* Deprecate the `type` flag value `Archive` and support `ZipArchive` instead.

`-type=Archive` (zip support) is effectively unused AFAICT, and could likely be removed entirely in a future release. Fetching archives of _either_ type is untested, which should also be addressed in a future cleanup PR.

cc @squee1945

### DIFF
--- a/gcs-fetcher/pkg/fetcher/fetcher.go
+++ b/gcs-fetcher/pkg/fetcher/fetcher.go
@@ -16,13 +16,16 @@ limitations under the License.
 package fetcher
 
 import (
+	"archive/tar"
 	"archive/zip"
+	"compress/gzip"
 	"context"
 	"crypto/sha1"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"math"
 	"math/rand"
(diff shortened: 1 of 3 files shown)


gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -651,23 +653,19 @@ func (gf *Fetcher) fetchFromManifest(ctx context.Context) (err error) {
...
+		if cerr := rc.Close(); cerr != nil {
+			err = fmt.Errorf("Failed to close file %q: %v", name, cerr)
 		}
 	}()
 
-	targetFile := filepath.Join(gf.DestDir, file.Name)
-	if err := gf.ensureFolders(targetFile); err != nil {
-		return fmt.Errorf("failed to create folders for file %q: %v", targetFile, err)
+	targetFile := filepath.Join(gf.DestDir, name)
+	if err := gf.OS.MkdirAll(filepath.Base(targetFile), mode); err != nil {
```
squee1945: Can you just use gf.DestDir here (instead of filepath.Base())?

ImJasonH: I don't think so. `gf.DestDir` is something like `/workspace`, while `name` might include a path to the file, e.g., `path/to/file.go`, so you need to `mkdir `/workspace/path/to/` before writing `file.go`. Does that match your understanding of this code?

squee1945:
LGTM

ImJasonH:
Added a test, which caught a bug, which I fixed! :tada:

gcs-fetcher/pkg/fetcher/fetcher.go
```diff
@@ -760,20 +762,123 @@ func (gf *Fetcher) fetchFromZip(ctx context.Context) (err error) {
...
+	}
+	gzr, err := gzip.NewReader(f)
+	if err != nil {
+		return err
+	}
+	tr := tar.NewReader(gzr)
+
+	defer func() {
+		if cerr := f.Close(); cerr != nil {
+			err = fmt.Errorf("Failed to close file %q: %v", tgzfile, cerr)
```
codrienne: Should this error be logged? Not sure where it gets used later.

ImJasonH: `fetchFromTarGz` uses a named return value, so this err gets returned by the method.

PR #544, opened 2019-08-06
[gcloud] Update Dockerfile.slim

MunGell:
Remove extra slash in PATH environment variable.

This is how PATH looks like in the image at the moment:

```bash
$ echo $PATH
/builder/google-cloud-sdk/bin/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

You can notice `/builder/google-cloud-sdk/bin/` is the only one that ends with `/`.

This results in following execution path:

```bash
$ which gcloud
/builder/google-cloud-sdk/bin//gcloud
```

Does not affect functionality. Cleanup PR.

### DIFF
--- a/gcloud/Dockerfile.slim
+++ b/gcloud/Dockerfile.slim
@@ -26,6 +26,6 @@ RUN apt-get -y update && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf ~/.config/gcloud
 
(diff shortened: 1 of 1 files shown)


bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #550, opened 2019-08-16
Switch to smaller images for small tools

mcrmonkey:
Switching to alpine from ubuntu to produce smaller images for utility
images wget and curl

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.9
 
-RUN apk --no-cache add curl
+RUN apk --no-cache add curl ca-certificates
 
 ENTRYPOINT ["curl"]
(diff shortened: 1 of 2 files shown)


mcrmonkey:
Using ubuntu for `wget` and `curl` images seemed to be a bit overkill for what was needed. Switched these to alpine ( pinned to 3.9 ) to produce a smaller image for what was needed.

ImJasonH:
Thanks for this contribution! Unfortunately some users of these images might have behavior that relys on the underlying OS, so changing from Ubuntu to Alpine might break them.

Furthermore, the choice of a [Google-managed base image](https://cloud.google.com/container-registry/docs/managed-base-images) was intentional, for many of the reasons listed at that page.

Luckily, these images are all cached on the VMs that run GCB builds, so the size of the image will have much less impact on users' build latencies.

PR #551, opened 2019-08-21
Add -X POST in automated deployments example.

joonlim:
Interestingly, the command works as it is, but adding -X POST works better (returns a response)


joonlim:
@ImJasonH Please review

PR #559, opened 2019-09-18
Not passing --namespace to gke-deploy allows configs to define namespace

joonlim:
- No --namespace flag means configs' namespace fields are not modified
- No --namespace flag means calling kubectl apply without a -n <namespace>
- If --namespace flag is supplied or if any namespace configs are passed, check if the namespace already exists. If it does not, warn the user and create the namespace.

To support this, each config is now applied individually through stdin.


joonlim:
@stewchen

joonlim:
@ImJasonH Could you please merge

PR #563, opened 2019-09-23
Update gke-deploy docs to suggest splitting prepare + apply.

joonlim:


joonlim:
@ImJasonH Could you please merge this?

PR #569, opened 2019-10-11
Gcloud fix

mafrosis:
A couple of trivial fixes/changes to the `gcloud` `Dockerfile`


ImJasonH:
Thanks for your contribution! Unfortunately I'm not sure we can safely remove `wget` from this image, since it's possible some user of the image depends on `wget` being included. 😞

mafrosis:
Very good point!

I can raise a PR with the other commit if you're interested in that. I didn't bother to figure out how the `--disable-installation-options` flag is continuing to work

PR #574, opened 2019-11-03
Add env var specifying kubectl version

jakebiesinger:
I want to use this builder and its nice ZONE, CLUSTER etc wrappers, but need more recent features (namely, `kubectl rollout restart`).

`gcloud` ships with several versions of `kubectl`, but the default is the somewhat old `1.13`. This PR seems like a happy medium, allowing folks to specify more recent versions that shipp w/gcloud without needing to build their own kubectl images. 

As of today, the `$versions` are 
- 1.13 (default)
- 1.14
- 1.15

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/cloud-builders/574)

### DIFF
--- a/kubectl/kubectl.bash
+++ b/kubectl/kubectl.bash
@@ -1,16 +1,21 @@
 #!/bin/bash
 
+versions=( $(cd "$(gcloud info --format='value(installation.sdk_root)')/bin/" && ls kubectl.* | sed s/kubectl.//g) )
+
 function var_usage() {
   cat <<EOF
   No cluster is set. To set the cluster (and the region/zone where it is found), set the environment variables
   CLOUDSDK_COMPUTE_REGION=<cluster region> (regional clusters)
   CLOUDSDK_COMPUTE_ZONE=<cluster zone>
   CLOUDSDK_CONTAINER_CLUSTER=<cluster name>
+
+  Optionally, you can specify the kubectl version via KUBECTL_VERSION, taking one of the following values: ${versions[@]}
 EOF
 
   exit 1
 }
 
+kubectl_cmd=kubectl.${KUBECTL_VERSION}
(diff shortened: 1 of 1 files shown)


jakebiesinger:
@googlebot I signed it!

ImJasonH:
Oh cool! I didn't know gcloud started shipping with multiple kubectl versions.

This looks fine to me. It might be annoying for a user that specifies `kubectl.1.14‘ when months later that version is no longer available, but the benefit to users wanting fresher releases balances it out imo.

kubectl/kubectl.bash
```diff
@@ -1,23 +1,32 @@
...
   CLOUDSDK_COMPUTE_ZONE=<cluster zone>
   CLOUDSDK_CONTAINER_CLUSTER=<cluster name>
+
+  Optionally, you can specify the kubectl version via KUBECTL_VERSION, taking one of the following values: ${versions[@]}
 EOF
 
   exit 1
 }
 
+kubectl_cmd=kubectl.${KUBECTL_VERSION}
```
camh-anz: This does not appear to be optional. If you do not specify `KUBECTL_VERSION`, it tries to run `kubectl.` and that doesn't exist. The period needs to only be added if `KUBECTL_VERSION` is set.

It appears this change has recently been pushed to `gcr.io` and is breaking our deployments. An urgent fix would be appreciated.

camh-anz: e.g. `kubectl_cmd=kubectl${KUBECTL_VERSION:+.${KUBECTL_VERSION}}`

jakebiesinger: oy. This is utter 🤦‍♂ and I'm so sorry. Let's get this fixed up ASAP

PR #576, opened 2019-11-08
Add a comment line to the created suggested YAML indicating that the container image will be replaced when in the expanded YAMLs.

joonlim:
Add a comment line to the created suggested YAML indicating that the container image will be replaced when in the expanded YAMLs.


joonlim:
@leeonlee @ImJasonH Could you please give this a review?

joonlim:
@codrienne

PR #578, opened 2019-11-11
Fix ~/.bazelrc not being used in Bazel builder

codesuki:
By using `--announce_rc` we found that the `.bazelrc` added in this Dockerfile was not being used when running in cloud build.
This results in steps not sharing the output base, i.e. being slow.

Under Cloud Build the home folder is `/builder/home` whereas under Docker it's `/root`. Therefore the `.bazelrc` is not found.


bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/667

PR #579, opened 2019-11-11
Override namespace by default and only add default namespace if object does not have namespace field.

joonlim:


joonlim:
@leeonlee @ImJasonH Could you please also give this a review?

joonlim:
@leeonlee @ImJasonH Could you also please give this a review?

joonlim:
@codrienne

gke-deploy/core/resource/resource.go
```diff
@@ -234,27 +234,30 @@ func UpdateMatchingContainerImage(ctx context.Context, objs Objects, imageName,
 }
 
 // UpdateNamespace updates all objects to change its namespace to the provided namespace. Objects
-// that do not have a namespace field will not be updated.
+// that do not have a namespace field will also be updated to have a namespace field.
 func UpdateNamespace(ctx context.Context, objs Objects, replace string) error {
-	var hasNS []*Object
+	for _, obj := range objs {
+		if err := setObjectNamespace(obj, replace); err != nil {
```
codrienne: so how does setObjectNamespace behave when the object has no namespace?

joonlim: setObjectNamespace will set the namespace field to an object whether or not they already have one. The previous implementation checked to make sure that the field existed before setting it, but due to suggestion from product, we are now always setting the namespace in the YAML if the user provides one.

PR #586, opened 2019-11-21
Unified apply file

gleeper:
Update `gke-deploy` to store all k8s resources in one aggregate file.

 - [Buganizer ticket](https://buganizer.corp.google.com/issues/133264279)

Note: `resource_test.go` and `deployer_test.go` passed locally.  I'm not aware of any other impacted tests.

## Changes Implemented

 - Changed `resource.Objects` from a map to a slice of Objects
 - Changed `parseResourceFromFile` to return an updated Objects, instead of modifying the one inputted
 - Removed `Objects.AddObject` (can just use append now)
 - Removed `fixedCollidingBaseNames`
 - Added `filteredObjs` to keep track of which objects should be deleted (or more specifically which ones should be kept).  This was needed to replace `delete(objs, filename)`

### DIFF
--- a/gke-deploy/core/resource/resource.go
+++ b/gke-deploy/core/resource/resource.go
@@ -165,6 +165,8 @@ func SaveAsConfigs(ctx context.Context, objs Objects, outputDir string, lineComm
 	aggregateName := "aggregate.yaml"
 	filename := filepath.Join(outputDir, aggregateName)
 
+	resources := []string{}
+
 	for _, obj := range objs {
 		out, err := runtime.Encode(encoder, obj)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


gleeper:
@joonlim

gleeper:
I'm wrapping up adding file diffs to the tests, so those changes will be out soon in a separate PR.

gke-deploy/core/resource/resource.go
```diff
@@ -159,8 +161,13 @@ func SaveAsConfigs(ctx context.Context, objs Objects, outputDir string, lineComm
 	if err := oss.MkdirAll(ctx, outputDir, os.ModePerm); err != nil {
 		return fmt.Errorf("failed to create output directory %q: %v", outputDir, err)
 	}
-	for baseName, obj := range objs {
-		filename := filepath.Join(outputDir, baseName)
+
+	aggregateName := "aggregate.yaml"
```
joonlim: Mentioned in person - maybe change this to "aggregated-resources.yaml"? If you choose to do so, make sure to fix up the tests that reference this as well. Might make sense to put it in a variable name that can be reference by the unit tests.

joonlim: @demobox FYI

demobox: > maybe change this to "aggregated-resources.yaml"

+1 to something like that. Or `kubernetes-resources.yaml`, `kubernetes-config.yaml`, `k8s-config.yaml` or so?

PR #588, opened 2019-11-22
Update gke-deploy tests to use actual files instead of stubbing os

gleeper:
Testing recursive directories via stubbing each os call would have gotten fairly convoluted rather quickly.  

As a precursor to implementing and, more relevantly, testing the recursive option on `gke-deploy`, I've changed the `ParseConfig` tests to use actual files instead of stubbing the `oss` input.  

`resource_test.go` (which is the only `go` file changed and, therefore, the only one needed to be tested) passed locally.


gleeper:
@joonlim request for your review please

PR #589, opened 2019-11-22
Refactor ParseConfigs as a precursor to adding the recursive flag to gke-deploy

gleeper:
Refactor `ParseConfigs` to walk through a directory (up to depth 1).

Adding the rest of the recursion logic (flag, tests, etc.) would make this PR a bit too bulky, so I'll save those changes for the next PR.  

I also added a couple tests around invalid yaml configs.

Cloud build was submitted and passed.

### DIFF
--- a/gke-deploy/core/resource/resource.go
+++ b/gke-deploy/core/resource/resource.go
@@ -91,47 +91,59 @@ func DecodeFromYAML(ctx context.Context, yaml []byte) (*Object, error) {
 func ParseConfigs(ctx context.Context, configs string, oss services.OSService) (Objects, error) {
 	objs := Objects{}
 
+	if configs == "-" {
+		objs, err := parseResourcesFromFile(ctx, configs, objs, oss)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


gleeper:
@joonlim request for review please.

gke-deploy/core/resource/resource.go
```diff
@@ -91,47 +91,59 @@ func DecodeFromYAML(ctx context.Context, yaml []byte) (*Object, error) {
...
 
-		hasResource := false
-		for _, fi := range files {
-			if fi.IsDir() {
-				continue
-			}
-			filename := filepath.Join(d, fi.Name())
-			if !hasYamlOrYmlSuffix(filename) {
-				continue
+	var walk func(path string, fi os.FileInfo,  depth int) error
```
joonlim: Not sure if this needs to be declared

gleeper: Unfortunately, it needs to be declared because it's recursive.  I'll add a comment here in my next PR

PR #591, opened 2019-11-25
Add recursive option to `resource.ParseConfigs`

gleeper:
I'm going to have a string of PRs to, in effect, bubble up the recursive flag.  This PR adds the recursive input and functionality to `ParseConfigs`.  All call sites (aside from tests), however, just set that input to false. 

I also changed `depth` in the `walk` function to a bool (`baseDir`) because we don't need to know exactly how deep we currently are, we only need to know if we are in the base directory or not.  

Cloud build was submitted and passed.


gleeper:

PR #592, opened 2019-11-25
Update gke-deploy deployer test to use actual files instead of stubbing the OS

gleeper:
Update `deployer/deployer_test.go` to use actual files when testing `Prepare`, as opposed to stubbing the OS.  This is a precursor to adding the recursive option to the deployer.

When testing `Prepare` errors, I also added a check on the actual error messages (when working with this, I once had all error tests pass because the `configsDir` didn't exist).  This will help to ensure that each set of conditions produces the proper error.

The next `PR` will make the same changes to the `Apply` tests.

`gcloud builds submit` executed successfully


gleeper:

PR #593, opened 2019-11-25
Update gke-deploy deployer.Apply tests to use actual files instead of stubbing the OS

gleeper:
In addition to the main change mentioned in the description, I also made the following changes in this PR:

 - Changed the default test timeout from 5 minutes to 10 seconds. The longest test, which checks the same resource twice, takes 5 seconds, so 10 should be plenty
 - Updated the test kubectl service to a list of (`Res`, `Err`) pairs (instead of a a pair of `Res` and `Err` lists).  This will ensure that we always specify the same number of responses and errors.
 - Used the new kubectl setup to verify that we `apply` and `get` the proper resources the proper number of times
 - Added logic to validate each failure condition produces the proper error

### DIFF
--- a/gke-deploy/core/cluster/cluster.go
+++ b/gke-deploy/core/cluster/cluster.go
@@ -30,7 +30,7 @@ func AuthorizeAccess(ctx context.Context, clusterName, clusterLocation, clusterP
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


gleeper:

gleeper:
Forgot to mention, `gcloud builds submit` execution succeeded

PR #594, opened 2019-11-26
Add recursive input to gke-deploy deployer.Prepare and Apply

gleeper:
In addition to adding the recursive argument, I also updated all call sites (in the `cmd` folder) of `Prepare` and `Apply` to set `recursive` to false.

Also added tests.

`gcloud builds submit` execution succeeded


gleeper:

PR #596, opened 2019-11-26
Only add to deployedObjs once the object is done deploying

gleeper:
As @joonlim pointed out, this was causing issues in the summary.  A line would be added to the summary every time a resource was checked, instead of just one line per resource. 

Moving the `append` into the `if ok` block fixes the issue.

`gcloud builds submit` execution succeeded.

### DIFF
--- a/gke-deploy/deployer/deployer.go
+++ b/gke-deploy/deployer/deployer.go
@@ -331,7 +331,8 @@ func (d *Deployer) Apply(ctx context.Context, clusterName, clusterLocation, clus
 		}
 	}
 
-	deployedObjs := resource.Objects{}
+	deployedObjs := map[string]map[string]resource.Object{}
+	summaryObjs := make(resource.Objects, 0, len(objs))
 	timedOut := false
 
 	fmt.Printf("\nWaiting for deployed objects to be ready with timeout of %v\n", waitTimeout)
@@ -364,12 +365,15 @@ func (d *Deployer) Apply(ctx context.Context, clusterName, clusterLocation, clus
 			if err != nil {
 				return fmt.Errorf("failed to get configuration of deployed object with kind %q and name %q: %v", kind, name, err)
 			}
+			if deployedObjs[kind] == nil {
+				deployedObjs[kind] = map[string]resource.Object{}
+			}
+			deployedObjs[kind][name] = *deployedObj
 			ok, err := resource.IsReady(ctx, deployedObj)
 			if err != nil {
 				return fmt.Errorf("failed to check if deployed object with kind %q and name %q is ready: %v", kind, name, err)
 			}
 			if ok {
-				deployedObjs = append(deployedObjs, deployedObj)
 				dur := time.Now().Sub(start).Round(time.Second / 10) // Round to nearest 0.1 seconds
 				fmt.Printf("Deployed object with kind %q and name %q is ready after %v\n", kind, name, dur)
 			} else {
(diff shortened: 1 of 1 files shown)


gleeper:

joonlim:
Nice work.

I noticed another bug that should be addressed here:

After a timeout, the deployment summary should display all objects, rather than only objects that are ready.

E.g.,
```
➜  gke-deploy run -i nginx -x 80 -t 10s              
Preparing deployment.
Starting with no configuration files
Creating suggested Deployment configuration file "nginx"
Creating suggested HorizontalPodAutoscaler configuration file "nginx-hpa"
Creating suggested Service configuration file "nginx-service"
Updating container of resource: {kind: Deployment, name: nginx}
Saving suggested configuration files to "output/suggested"

Expanding configuration files.
Got digest for image: index.docker.io/library/nginx:latest --> index.docker.io/library/nginx@sha256:189cce606b29fb2a33ebc2fcecfa8e33b0b99740da4737133cdbcee92f3aba0a
Updating containers in configuration files that have image name "index.docker.io/library/nginx" to use image with digest "index.docker.io/library/nginx@sha256:189cce606b29fb2a33ebc2fcecfa8e33b0b99740da4737133cdbcee92f3aba0a"
Updating container of resource: {kind: Deployment, name: nginx}
Saving expanded configuration files to "output/expanded"
Finished preparing deployment.

Applying deployment.
Configuration files to be used: [{kind: Deployment, name: nginx} {kind: HorizontalPodAutoscaler, name: nginx-hpa} {kind: Service, name: nginx-service}]
Applying configuration files to cluster.

Waiting for deployed objects to be ready with timeout of 10s
Deployed object with kind "HorizontalPodAutoscaler" and name "nginx-hpa" is ready after 400ms
Deployed object with kind "Deployment" and name "nginx" is ready after 5.2s
Finished applying deployment.

################################################################################
> Deployed Objects

NAMESPACE    KIND                       NAME         READY    
default      Deployment                 nginx        Yes      
default      HorizontalPodAutoscaler    nginx-hpa    Yes      

################################################################################
> GKE

Workloads:             https://console.cloud.google.com/kubernetes/workload?project=cbd-test
Services & Ingress:    https://console.cloud.google.com/kubernetes/discovery?project=cbd-test
Applications:          https://console.cloud.google.com/kubernetes/applications?project=cbd-test
Configuration:         https://console.cloud.google.com/kubernetes/config?project=cbd-test
Storage:               https://console.cloud.google.com/kubernetes/storage?project=cbd-test

Error: failed to apply deployment: timed out after 10s while waiting for deployed objects to be ready
```

I should see a line that indicates that the nginx-service is not ready.
This worked before because it was easier to replace items in the map: https://github.com/GoogleCloudPlatform/cloud-builders/commit/0ac7efef00d1cceef48911c39486157e9bbaeec9#diff-6a6b5c8660db21efbf6922b1d8a8c10eL367

gleeper:
Ah right.  Okay, I'll probably just change `deployedObjs` back to using a map.  No real need for that to use a list I suppose.

joonlim:
> p

I believe you can use a map that maps the objects to true (go's set alternative): https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure

gleeper:
In case `GetDeployedObject` returns differently hashed objects (because the underlying yaml is different even when pointing to the same resource), I changed `deployedObjs` to a two layer map from `kind (a string)` to `name (a string)` to `resource.Object`.  I think that should be sufficient to ensure that each k8s resource is accounted for exactly once in the summary regardless of status.  Let me know what you think when you get a chance, and thanks again for the comments!

PR #597, opened 2019-12-03
Fix duplicate entry issue in gke-deploy summary

gleeper:
When `range` iterates over a map, the `value` it creates is just a pointer.  So, we were initially assigning the same pointer for all objects with the same kind (because this only pertains to the second `range` loop).  

`gcloud builds submit` execution passed.


gleeper:

PR #599, opened 2019-12-16
Explicitly include empty-directory to fix unit tests

gleeper:
The `empty-directory` directory is required for tests, but wasn't included by git because it was an empty directory.  Adding a blank `.gitignore` file allows the empty directory to be included.


gleeper:
@joonlim

PR #602, opened 2020-01-03
Update gke-deploy docs with -R|--recursive flag.

joonlim:


joonlim:
@gleeper Please review

PR #611, opened 2020-01-22
Instructions for dockerhub gradle image

loosebazooka:
This seems to work for me despite the comments in: https://github.com/GoogleCloudPlatform/cloud-builders/pull/452/files#diff-5f83fe68243f99b18b871a104e9c86faR7-R12

Perhaps the way they build changed, and this now works? I don't know.

works for `gradle:`
- `6.0.1-jdk8`
- `6.0.1-jdk11` 
- `5.6.2-jdk8` 
- `5.6.2-jdk11`

### DIFF
--- a/gradle/README.md
+++ b/gradle/README.md
@@ -2,6 +2,17 @@
 
 This Cloud Build builder runs the Gradle build tool.
 
+You should consider instead using an [official `gradle`
+image](https://hub.docker.com/_/gradle/) and specifying the `gradle` entrypoint:
+```yaml
+steps:
+- name: gradle:6.0.1-jdk11
+  entrypoint: 'gradle'
+  args: ['build']
+```
+This allows you to use any supported version of Gradle with any supported JDK
+version.
+
 ## Building this builder
 
 To build this builder, run the following command in this directory.


chanseokoh:
I'd like to verify this myself. Let me do some test.

chanseokoh:
Ah, yeah, looks like the official images changed the default user to root.
```
$ docker run --rm -it --entrypoint whoami gradle:4.6-jdk8
gradle
$ docker run --rm -it --entrypoint whoami gradle
root
```

chanseokoh:
Perhaps GCB should update Gradle versions in #452 and merge it so that GCB uses the official Gradle images going forward. I believe GCB already did this for Maven.

gradle/README.md
```diff
@@ -2,6 +2,17 @@
 
 This Cloud Build builder runs the Gradle build tool.
 
+You should consider instead using an [official `gradle`
+image](https://hub.docker.com/_/gradle/) and specifying the `gradle` entrypoint:
+```yaml
+steps:
+- name: gradle:6.0.1-jdk11
+  entrypoint: 'gradle'
+  args: ['build']
```
chanseokoh: And this should be `['gradle', 'build']` for `latest`. I see the entrypoint is null.

loosebazooka: I think as instructions to users it's fine? Matches the maven instructions, and safely overrides any potential entrypoint that they might set?

chanseokoh: Ah never mind, I didn't see `entrypoint` above.

bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/665

PR #612, opened 2020-01-23
kubectl: gcloud crashes if HTTPS_PROXY is set, fix it

IlariExove:
I need to set HTTPS_PROXY environment variable for kubectl but unfortunately gcloud command crashes if HTTPS_PROXY is set. This change unsets HTTPS_PROXY env variable before invoking gcloud commands and restores it before invoking kubectl commands.

My use case is a private GKE cluster that has master authorized networks fixed to certain static IP addresses. I run authenticating Tinyproxy in the GKE cluster as a container to allow Cloud Build instances to use kubectl as if it were from master authorized networks.

### DIFF
--- a/kubectl/kubectl.bash
+++ b/kubectl/kubectl.bash
@@ -28,13 +28,26 @@ if [[ -n "$KUBECTL_VERSION" ]] && [[ ! " ${versions[*]} "  =~ " ${KUBECTL_VERSIO
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


bendory:
closing as obsolete in the context of
https://github.com/GoogleCloudPlatform/cloud-builders/pull/675
https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #620, opened 2020-02-07
Add GCS service for copying files between local disk and GCS.

ChrisGe4:
@joonlim @brandonnelson3

### DIFF
--- a/gke-deploy/core/gcs/gcs.go
+++ b/gke-deploy/core/gcs/gcs.go
@@ -5,35 +5,38 @@ import (
 	"context"
 	"errors"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/cloud-builders/gke-deploy/services"
 )
 
-const TimeoutErr = "GCS timeout"
+const (
+	timeoutErr = "GCS timeout"
+)
 
 var (
 	defaultTimeout = 1 * time.Hour
 	defaultDelay   = 2 * time.Second
-	errGCSTimeout  = errors.New(TimeoutErr)
+	errGCSTimeout  = errors.New(timeoutErr)
 )
 
 type GCS struct {
-	timeout    time.Duration
-	retries    int
-	delay      time.Duration
-	gcsService services.GcsService
+	Timeout    time.Duration
+	Retries    int
+	Delay      time.Duration
+	GcsService services.GcsService
 }
 
 // Download copies file(s) from GCS. <dst> should be a directory, not a path to a file
 func (s *GCS) Download(ctx context.Context, src, dst string, recursive bool) error {
-	log.Printf("Downloading file(s) from GCS. Source: %s  Destination: %s.\n", src, dst)
+	log.Printf("Downloading file(s) %s from %s.\n", src, dst)
 	return s.copyWithRetry(ctx, src, dst, recursive)
 }
 
 // Upload copies file(s) to GCS.
 func (s *GCS) Upload(ctx context.Context, src, dst string) error {
-	log.Printf("Uploading file(s) to GCS.  Source: %s  Destination: %s.\n", src, dst)
+	log.Printf("Uploading file(s) %s to %s.\n", src, dst)
 	return s.copyWithRetry(ctx, src, dst, false)
 
 }
@@ -42,18 +45,21 @@ func (s *GCS) Upload(ctx context.Context, src, dst string) error {
 // with appropriate retry backoff.
 func (s *GCS) copyWithRetry(ctx context.Context, src, dst string, recursive bool) error {
 	var err error
-	delay := s.delay
+	delay := s.Delay
 	if int64(delay) == 0 {
 		delay = defaultDelay
 	}
-	for retryNum := 0; retryNum <= s.retries; retryNum++ {
+	for retryNum := 0; retryNum <= s.Retries; retryNum++ {
 		if retryNum > 0 {
 			time.Sleep(delay)
 		}
-		timeout := s.getTimeout()
+		timeout := s.timeout()
 		e := s.copyWithTimeout(ctx, src, dst, recursive, timeout)
 		if e != nil {
 			err = e
+			if strings.Contains(err.Error(), "AccessDeniedException") {
+				return err
+			}
 			continue
 		}
 		return nil
@@ -66,29 +72,20 @@ func (s *GCS) copyWithRetry(ctx context.Context, src, dst string, recursive bool
 // copyWithTimeout calls GcsService to move the files. The call will time out if it
 // takes too long.
 func (s *GCS) copyWithTimeout(ctx context.Context, src, dst string, recursive bool, timeout time.Duration) error {
-	status := make(chan error, 1)
-	go func() {
-		status <- s.gcsService.Copy(ctx, src, dst, recursive)
-	}()
-
-	select {
-	case <-ctx.Done():
-		if ctx.Err() == context.DeadlineExceeded {
-			return errGCSTimeout
-		}
-		return ctx.Err()
-	case <-time.After(timeout):
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	err := s.GcsService.Copy(ctx, src, dst, recursive)
+	if ctx.Err() == context.DeadlineExceeded {
 		return errGCSTimeout
-	case err := <-status:
-		return err
 	}
+	return err
 
 }
 
 //timeout returns the GCS timeout that will be used to call GcsService.
-func (s *GCS) getTimeout() time.Duration {
-	if int64(s.timeout) == 0 {
+func (s *GCS) timeout() time.Duration {
+	if int64(s.Timeout) == 0 {
 		return defaultTimeout
 	}
-	return s.timeout
+	return s.Timeout
 }
--- a/gke-deploy/core/gcs/gcs.go
+++ b/gke-deploy/core/gcs/gcs.go
@@ -1,15 +1,3 @@
-/*
-Copyright 2019 Google, Inc. All rights reserved.
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-    http://www.apache.org/licenses/LICENSE-2.0
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


ChrisGe4:
@googlebot I signed it!

ChrisGe4:
@googlebot I signed it!

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
+/*
```
joonlim: Remove the license

gke-deploy/testservices/gsutil.go
```diff
@@ -0,0 +1,26 @@
+package testservices
+
+import (
+	"context"
+	"fmt"
+)
+
+// GCSResponse simulates errors while calling GcsService.
```
joonlim: remove?

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+type GCS struct {
+	Timeout    time.Duration
+	Retries    int
+	Delay      time.Duration
+	GcsService services.GcsService
+}
+
+// Download copies file(s) from GCS. <dst> should be a directory, not a path to a file
+func (s *GCS) Download(ctx context.Context, src, dst string, recursive bool) error {
+	log.Printf("Downlaoding file(s) from GCS. Source: %s  Destination: %s.\n", src, dst)
```
joonlim: Downlaoding -> Downloading

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+
+const TimeoutErr = "GCS timeout"
+
+var (
+	defaultTimeout = 1 * time.Hour
+	errGCSTimeout  = errors.New(TimeoutErr)
+)
+
+type GCS struct {
+	Timeout    time.Duration
```
joonlim: Can these fields be private or do you plan to use them outside of this package?

gke-deploy/core/gcs/gcs_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+	gcs := buildTestGCS(t)
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+
+			err := gcs.Download(context.Background(), tc.src, tc.dst, tc.recursive)
+			if (err == nil) != tc.ok {
+				t.Errorf("Test should %s but not", tc.status)
+				return
```
joonlim: lowercase "test"

gke-deploy/core/gcs/gcs_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	gcs := buildTestGCS(t)
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+
+			err := gcs.Download(context.Background(), tc.src, tc.dst, tc.recursive)
+			if (err == nil) != tc.ok {
+				t.Errorf("Test should %s but not", tc.status)
```
joonlim: This codebase tries to follow a pattern of describing what args were passed in for the failing case: e.g., https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/gke-deploy/core/cluster/cluster_test.go#L27

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+		started := time.Now()
+		timeout := s.timeout()
+		e := s.copyWithTimeout(ctx, src, dst, recursive, timeout)
+		if e != nil {
+			err = e
+			log.Printf("Started copying at %v and failed at %v.", started, time.Now())
+			continue
+		}
+		log.Printf("Started copying at %v and finished at %v.", started, time.Now())
+		break
```
joonlim: I think there is a bug here: If you fail the first time and succeed the second time, wouldn't err be set to the first error and be returned?

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return err
+
+}
+
+// copyWithTimeout calls GcsService to move the files. The call will time out if it
+// takes too long.
+func (s *GCS) copyWithTimeout(ctx context.Context, src, dst string, recursive bool, timeout time.Duration) error {
+	status := make(chan error, 1)
+	log.Printf("Operation will time out in %f seconds", timeout.Seconds())
```
joonlim: Question about these log statements. Do these get logged to a file or to stdout? I'm not sure if we want the user to see all this info in stdout

gke-deploy/core/gcs/gcs.go
ChrisGe4: My idea is those variable should be changeable after GCS instance was created. E.g, Retries could be modify dynamically if the failure rate is high in certain time. But it is debatable since we don't have such requirements in this project. Do you think it is better to hide them?

brandonnelson3: If we don't expect to need to change them I would avoid "future proofing". You should always minimize visibility where possible. In the future if that functionality is needed, we can expose it then.

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+// copyWithRetry is responsible for trying (and retrying) to call copyWithTimeout()
+// with appropriate retry backoff.
+func (s *GCS) copyWithRetry(ctx context.Context, src, dst string, recursive bool) error {
+	var err error
+	delay := s.Delay
+	for retryNum := 0; retryNum <= s.Retries; retryNum++ {
+
+		if retryNum > 0 {
+			time.Sleep(delay)
+			delay *= 2
```
brandonnelson3: It seems to me like just a fixed delay would be sufficient here, instead of this exponential backoff since most of the time I would guess network issues are the cause of any temporary errors here. Also worth noting that a time.Duration is just an int64 under the hood so the default is 0 and this will infinitely loop. https://golang.org/pkg/time/#Duration

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		started := time.Now()
+		timeout := s.timeout()
+		e := s.copyWithTimeout(ctx, src, dst, recursive, timeout)
+		if e != nil {
```
brandonnelson3: I would also check this error to see if it is a permanent error like a not found error or something, no point retrying if the dst is a bucket that doesn't exist for instance.

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,109 @@
...
+			err = e
+			log.Printf("Started copying at %v and failed at %v.", started, time.Now())
```
brandonnelson3: This log is going to be very hard to read:

Started copying at 2009-11-10 23:00:00 +0000 UTC m=+0.000000001 and failed at 2009-11-10 23:00:00 +0000 UTC m=+0.000000001.

https://play.golang.org/p/Gxmd_yyxGHM

Consider formatting it with something from https://gobyexample.com/time-formatting-parsing

gke-deploy/core/gcs/gcs.go
brandonnelson3: You can print time.Durations with %s more cleanly:

https://play.golang.org/p/o0GemUyv1qN

gke-deploy/core/gcs/gcs.go
ChrisGe4: Done

gke-deploy/core/gcs/gcs_test.go
ChrisGe4: Add a test for errors.

gke-deploy/core/gcs/gcs.go
ChrisGe4: Done.

gke-deploy/core/gcs/gcs.go
ChrisGe4: Good catch! I fixed the bug and added a new test case.

gke-deploy/core/gcs/gcs.go
ChrisGe4: discussed offline.

ChrisGe4: Good to know.

gke-deploy/core/gcs/gcs.go
ChrisGe4: Discussed with Joon, seems like it is better to not have these logs. I removed them.

gke-deploy/core/gcs/gcs.go
ChrisGe4: I made the change to use a fixed delay. I don't quite get the infinite loop part though.

gke-deploy/core/gcs/gcs.go
ChrisGe4: runCommand doesn't return the error in detail. it will merely return the status code.

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,91 @@
...
+	return s.copyWithRetry(ctx, src, dst, false)
+
+}
+
+// copyWithRetry is responsible for trying (and retrying) to call copyWithTimeout()
+// with appropriate retry backoff.
+func (s *GCS) copyWithRetry(ctx context.Context, src, dst string, recursive bool) error {
+	var err error
+	delay := s.Delay
+	if int64(delay) == 0 {
```
joonlim: I think Brandon mentioned that you don't have to cast this to `int64` since its already an alias of `int64`: https://golang.org/pkg/time/#Duration

gke-deploy/core/gcs/gcs.go
```diff
@@ -0,0 +1,91 @@
...
+	defaultTimeout = 1 * time.Hour
+	defaultDelay   = 2 * time.Second
+	errGCSTimeout  = errors.New(timeoutErr)
+)
+
+type GCS struct {
+	Timeout    time.Duration
+	Retries    int
+	Delay      time.Duration
+	GcsService services.GcsService
```
joonlim: Did you intentionally make these public again?

ChrisGe4: Yes. I feel that it makes more sense to me to keep them public. It makes this class easier to use. I'm ok to make a change if you guys are strongly against this pattern.

PR #631, opened 2020-02-24
Ensure Application CRD is installed when attempting to apply an Application CR.

joonlim:


joonlim:
@ChrisGe4 Please review

gke-deploy/services/kubectl.go
```diff
@@ -37,7 +49,10 @@ func (k *Kubectl) ApplyFromString(ctx context.Context, configString, namespace s
 
 // Get calls `kubectl get <kind> <name> -n <namespace> --output=<format>`.
 func (k *Kubectl) Get(ctx context.Context, kind, name, namespace, format string, ignoreNotFound bool) (string, error) {
-	args := []string{"get", kind, name}
+	args := []string{"get", kind}
```
ChrisGe4: what will happen if name is empty？

gke-deploy/core/crd/crd.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// EnsureInstallApplicationCRD ensures the installation of the Application CRD in the current context's cluster.
+func EnsureInstallApplicationCRD(ctx context.Context, ks services.KubectlService) error {
+	installed, err := crdIsInstalled(ctx, applicationCRDName, ks)
+	if err != nil {
+		return err
+	}
+	if installed {
+		return nil
```
ChrisGe4: Should we have a log here if the crd has been installed?

gke-deploy/services/kubectl.go
joonlim: Good catch. I should add a validation, especially since we don't test these services.

gke-deploy/core/crd/crd.go
joonlim: I didn't feel like it would be useful. I want the installation of the CRD to be transparent

gke-deploy/services/kubectl.go
joonlim: On second thought, there is no validation to be done here.

If name is empty, it just called kubectl get <kind>

The reason why one would want to do this is to

1. Get a list of all resources of one kind
```
➜  kubectl get deployment                                                                   
NAME              READY   UP-TO-DATE   AVAILABLE   AGE
test-deployment   1/1     1            1           34d
```

2. Or in this case, check that the Application CRD is installed
```
➜  kubectl get customresourcedefinition.apiextensions.k8s.io/applications.app.k8s.io          
NAME                      CREATED AT
applications.app.k8s.io   2020-02-21T19:58:14Z
```

ChrisGe4:
LGTM

PR #636, opened 2020-03-06
--annotations and --links flags can take key/value pairs where the value has '=' signs.

joonlim:


gke-deploy/cmd/common/common.go
```diff
@@ -25,9 +25,9 @@ func CreateApplicationLinksListFromEqualDelimitedStrings(applicationLinks []stri
 		if p == "" {
 			continue
 		}
-		kv := strings.Split(p, "=")
+		kv := strings.SplitN(p, "=", 2)
 		if len(kv) != 2 {
-			return nil, fmt.Errorf("invalid key value pair: %q", p)
+			return nil, fmt.Errorf("key value pair must be separated by a '=' character")
```
louisjimenez: would it be helpful to also display the invalid key value pair with the error message?

Issue #638, opened 2020-03-22
Deprecate cloud-builders/gcloud in favor of google/cloud-sdk on DockerHub


ludoch:
Please, update the cloud-builders/gcloud/ image with a newer version of the Cloud SDK

The image contains the obsolete version SDK 279.0.0 and it should above 284 now.,

This is blocking GCP customers.
Maybe the image should be updated every where, like the Cloud SDK itself, or the image should do a gcloud components update before running a command.

Leejjon:
With a cloudbuild.yaml file that contains:

steps:
  - name: 'gcr.io/cloud-builders/gcloud'
    args: ['--version']

I managed to run the 'gcloud --version' command in cloud build:

![gcloud--version](https://user-images.githubusercontent.com/1529142/77254339-c5d5ba00-6c60-11ea-8290-f16b1c9580a9.png)


But as you can see it's still on SDK 279.0.0, so running it with a cloudbuild.yaml that uses one of the new beta commands:

steps:
  - name: 'gcr.io/cloud-builders/gcloud'
    args: ['beta app deploy pom.xml']

Doesn't work:
![gcloudbeta](https://user-images.githubusercontent.com/1529142/77254416-28c75100-6c61-11ea-9d14-c3a01e4afeb2.png)

irvifa:
I believe the image already download the latest:
```
FROM launcher.gcr.io/google/ubuntu16_04

RUN apt-get -y update && \
    apt-get -y install gcc python2.7 python-dev python-setuptools wget ca-certificates \
       # These are necessary for add-apt-respository
       software-properties-common python-software-properties && \

    # Install Git >2.0.1
    add-apt-repository ppa:git-core/ppa && \
    apt-get -y update && \
    apt-get -y install git && \

    # Setup Google Cloud SDK (latest)
    mkdir -p /builder && \
    wget -qO- https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz | tar zxv -C /builder && \
    CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
        --bash-completion=false \
        --disable-installation-options && \
    # install crcmod: https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod
    easy_install -U pip && \
    pip install -U crcmod && \

    # Clean up
    apt-get -y remove gcc python-dev python-setuptools wget && \
[...]

PR #643, opened 2020-04-06
gcloud-slim: remove apt cache

immerrr:
Looks like there's ~170Mb wasted in gcloud-slim image because apt cache is not cleaned before exiting:
image

I wonder if another 150Mb can be shaved off gcloud-slim by removing `/builder/google-cloud-sdk/.install/.backup` dir:
image

### DIFF
--- a/gcloud/Dockerfile
+++ b/gcloud/Dockerfile
@@ -28,4 +28,6 @@ RUN apt-get -y update && \
     /builder/google-cloud-sdk/bin/gcloud components list && \
 
     # Clean up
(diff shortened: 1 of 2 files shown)


immerrr:
I have also added commands to clean up the non-essential apt packages after the installation, and that saves another ~120Mb and brings the `gcloud-slim` primary layer size down from 1Gb to 750Mb.

bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #644, opened 2020-04-07
docker: don't install recommends except pigz, clean up apt afterwards

immerrr:
This PR slims down the last layer of `docker` image from 823MB to 365MB by doing the following:
- clean up apt caches and package lists
- remove software-properties-common that was only necessary when adding the docker repo
- not installing recommended packages (with pigz being notable exception to speed up layer decompression)

image


bendory:
Thank you for this PR; closing as obsolete in the context of https://github.com/GoogleCloudPlatform/cloud-builders/issues/670

PR #650, opened 2020-04-19
Allowing custom kube config in kubectl cloud builder

udatny:
this change would allow to provide a custom kube configuration and avoid gcloud to try to fetch credentials and do other fancy things. 

an example cloudbuild.yml related to this change might look like: 

```
steps:
## copy the kube config from gs storage
- name: gcr.io/cloud-builders/gsutil
  args: ['cp', 'gs://whateverbucket/mykubeconfig.yml', '.']

# This step uses the KUBERNETES_CONFIG_FILE option and just outputs the kube config
- name: 'gcr.io/cloud-builders/kubectl'
  id: ShowConfig
  args:
  - 'config'
  - 'view'
  env:
  - 'KUBERNETES_CONFIG_FILE=mykubeconfig.yml'

```

### DIFF
--- a/kubectl/kubectl.bash
+++ b/kubectl/kubectl.bash
@@ -8,7 +8,7 @@ function var_usage() {
   CLOUDSDK_COMPUTE_REGION=<cluster region> (regional clusters)
   CLOUDSDK_COMPUTE_ZONE=<cluster zone>
   CLOUDSDK_CONTAINER_CLUSTER=<cluster name>
-
+  KUBERNETES_CONFIG_FILE=path to alternate kubectl config in your workspace
   Optionally, you can specify the kubectl version via KUBECTL_VERSION, taking one of the following values: ${versions[@]}
 EOF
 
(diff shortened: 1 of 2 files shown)


udatny:
@googlebot I signed it!

kubectl/kubectl.bash
```diff
@@ -8,33 +8,42 @@ function var_usage() {
...
-fi
 
-if [ -n "$region" ]; then
-  echo "Running: gcloud container clusters get-credentials --project=\"$project\" --region=\"$region\" \"$cluster\""
-  gcloud container clusters get-credentials --project="$project" --region="$region" "$cluster" || exit
-else
-  echo "Running: gcloud container clusters get-credentials --project=\"$project\" --zone=\"$zone\" \"$cluster\""
-  gcloud container clusters get-credentials --project="$project" --zone="$zone" "$cluster" || exit
- fi
+if [ -z "$KUBERNETES_CONFIG_FILE" ]; then
```
irvifa, line 20:
```suggestion
```
I think we still need to fill it even though we already provided the custom kubeconfig aren't we?

kubectl/kubectl.bash
```diff
@@ -8,33 +8,42 @@ function var_usage() {
...
+  fi
+
+  if [ -n "$region" ]; then
+    echo "Running: gcloud container clusters get-credentials --project=\"$project\" --region=\"$region\" \"$cluster\""
+    gcloud container clusters get-credentials --project="$project" --region="$region" "$cluster" || exit
+  else
+    echo "Running: gcloud container clusters get-credentials --project=\"$project\" --zone=\"$zone\" \"$cluster\""
+    gcloud container clusters get-credentials --project="$project" --zone="$zone" "$cluster" || exit
+   fi
+fi
```
irvifa, line 40:
```suggestion
```

kubectl/kubectl.bash
udatny: the idea here is, that a completely google-independent k8s cluster could be addressed. so, no, we dont need to fill gcloud params in such a case. maybe the case you have on your mind also exists... but i didnt think about such one yet. the proposed change targets remote clusters, independent from gcloud.

bendory:
closing as obsolete; reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/638

PR #657, opened 2020-05-07
[npm] Build more versions and in parallel

naseemkullah:
* Adds `lts` and `12` to versions built
* makes `latest` actually latest and `current` actually current.
* Adds an id to each step and allow the builds to happen in parallel by means of waitFor.

Signed-off-by: Naseem <naseem@transit.app>


bendory:
Closing as obsolete, reference https://github.com/GoogleCloudPlatform/cloud-builders/issues/322

PR #673, opened 2020-05-15
Added deprecation notice for curl builder.

bendory:
https://github.com/GoogleCloudPlatform/cloud-builders/issues/672

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,3 +1,3 @@
-FROM marketplace.gcr.io/google/ubuntu1604
+FROM launcher.gcr.io/google/ubuntu1604
 COPY deprecation.sh /usr/bin
 ENTRYPOINT ["/usr/bin/deprecation.sh"]
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,7 +1,3 @@
-FROM launcher.gcr.io/google/ubuntu16_04
-
-RUN apt-get update && \
-  apt-get -y install curl ca-certificates
-
+FROM marketplace.gcr.io/google/ubuntu1604
 COPY deprecation.sh /usr/bin
 ENTRYPOINT ["/usr/bin/deprecation.sh"]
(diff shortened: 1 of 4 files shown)


curl/README.md
```diff
@@ -16,7 +17,7 @@ file must be publicly readable, since no credentials are passed in the request.
 
 ```
 steps:
-- name: gcr.io/cloud-builders/curl
+- name: 'curlimages/curl'
   args: ['http://www.example.com/remotefile.zip', '--output', 'localfile.zip']
```
leeonlee, line 21: This doesn't work because the new curl image runs as a non root user and can't write locally

bendory: ewww...

I'll update the example to one that works...

curl/Dockerfile
```diff
@@ -1,6 +1,3 @@
-FROM launcher.gcr.io/google/ubuntu16_04
-
-RUN apt-get update && \
```
bendory, line 3: The base image already has both `curl` and `ca-certificates` installed.

curl/README.md
```diff
@@ -1,9 +1,21 @@
...
-Arguments passed to this builder will be passed to `curl` directly.
+This image is deprecated.
+
+For best support of `curl` please use one of the official `curl` images
+maintained by the `curlimages` community on Dockerhub.
+
+For details, visit https://hub.docker.com/r/curlimages/curl.
+
+Note that `curlimages/curl` executes as special user `curl_user` and thus may
+not echo be suitable for all purposes.
```
sanastos, line 13: `not echo be suitable`

bendory: good catch, thanks!

curl/deprecation.sh
```diff
@@ -0,0 +1,22 @@
...
+echo           \*\*\*\*\* DEPRECATION NOTICE \*\*\*\*\*
+echo
+echo This image is deprecated and will no longer be updated.
+echo This recent version of the image will continue to exist.
+echo
+echo For best support of \`curl\` please use one of the official
+echo \`curl\` images maintained by the curlimages community
+echo on Dockerhub.
+echo For details, visit https://hub.docker.com/r/curlimages/curl. Note that
+echo this image executes as special user \`curl_user\` and thus may not echo be
```
leeonlee, line 12: There's an 'echo' out of place in this line

bendory: thanks, fixed

PR #683, opened 2020-05-26
Install matching Docker CLI versions for javac and bazel

leeonlee:
The newest (19.03.9) Docker CLI was being installed and due to https://github.com/docker/cli/issues/2533, it was not compatible with the daemon currently running in production (18.09)


bendory:
LGTM

Issue #677, opened 2020-05-18
Promote `golang` usage for `go` users


bendory:

cagedmantis:
The README.md document states that [golang](https://hub.docker.com/_/golang) `is provided by the Go team`. It may give the impression that the Go team maintains the docker images that are linked to. I think it would be helpful to clarify this.

haroonc:
Will be covered under upcoming update for cloud-builders

Issue #670, opened 2020-05-15
Promote https://hub.docker.com/_/docker images for `docker` users


bendory:

killthekitten:
@bendory is there a press release or any note regarding #680? Is it still a deprecation? 

The language has changed from deprecation to promoting official images, but it's not clear what consequences does it have for the end user.

bendory:
There is no deprecation. We have an internal plan to offer better versions of the `cloud-builders` that will offer better support, and we also plan to continue to notify users of the community-supported images as alternative choices.

Once we have full agreement + all the appropriate internal approvers, the main README will be updated to clarify the plans.

killthekitten:
Makes sense, thanks for the clarification!

bendory:
Of course -- and thank you for keeping us honest by asking! You'll note that I updated the main README with a "[Future Direction](https://github.com/GoogleCloudPlatform/cloud-builders/#future-direction)" section.

killthekitten:
Yes, noted 👍 I'm excited to move to official repos, thanks for the great work!

Issue #667, opened 2020-05-14
Promote `gcr.io/cloud-marketplace-containers/google/bazel` for `bazel` users


bendory:

ImJasonH:
You might be able to just merge https://github.com/GoogleCloudPlatform/cloud-builders/pull/504, if nobody is still using the special-case logic.

bendory:
> You might be able to just merge #504, if nobody is still using the special-case logic.

Not an unreasonable approach, but rather than try to chase that down I'm going to follow the same approach I'm using in other deprecations:

- Add a deprecation notice pointing users to an official replacement
- Stop building the image / stop making updates
- Leave the existing built image in place so that current users are not broken if they choose not to move

kyessenov:
Can we install tzdata into bazel image please? `/usr/share/zoneinfo` is missing in the xenial image.

wallrj:
In the cert-manager release tooling we'd like to use `gcr.io/cloud-marketplace-containers/google/bazel` but  we wanted a Bazel 4.2.1 image and couldn't find one, so we switched to `gcr.io/cloud-builders/bazel`.
Unfortunately, that Docker image doesn't have any tags, only digests:
 * https://github.com/cert-manager/release/pull/47

haroonc:
It seems `gcr.io/cloud-marketplace-containers/google/bazel` is not updated. Therefore it does not seem appropriate to deprecate this builder image in favour of marketplace image.

Issue #665, opened 2020-05-14
Promote community-supported builders for java toolchain users


bendory:
- Deprecate `mvn` builder in favor of https://hub.docker.com/_/maven
- Deprecate `gradle` builder in favor of https://hub.docker.com/_/gradle
- Deprecate `javac` builder in favor of https://hub.docker.com/_/openjdk

chanseokoh:
See #452 for Gradle.

PR #686, opened 2020-05-28
Fix argument forwarding in notice entrypoints.

rafikk:
Arguments should be forwarded quoted.


bendory:
Context:

* #677 
* #672 
* #670 
* #667 
* #665 
* #663 
* #638 
* #322

PR #694, opened 2020-06-03
Configure git in cloud-sdk to use ADC.

bendory:
This PR moves configuration of Application Default Credentials up from the `git` builder into the `gcloud` builders. Now ADC will be used for `git` in any image derived from `gcloud:slim`.

Also: document git functionality for all `cloud-sdk` variants.


bendory:
Context: #676

PR #714, opened 2020-08-06
Move all notices to bash.

bendory:
Sometimes `/bin/sh` doesn't like $RANDOM -- it depends on the underlying choice of `sh`. `bash` always has it.


bendory:
Rejecting in favor of already merged #715

PR #721, opened 2020-09-12
NPM & Yarn :: alpine versions

eeertekin:


bendory:
Can you please provide some context around why you would like to see GCB support `alpine`-based versions of these builders? What do the `alpine` versions provide that the existing versions don't?

eeertekin:
Hey @bendory,

Alpine version is around 350+ MB while Debian version is almost 1 GB. There is no additional benefits except the image size, but it directly affects the build times.

Simple hello world app without any additional module takes around 90-120 secs on Debian version (these images are not cached on GCB, on each build, they have to be pulled first) Same app with alpine version is around 30-60 secs.

If you use an old version of nodejs (8.x), since these old versions are already cached on GCB, build times decrease from 60+ secs to 10-15 secs.

Docker HUB version of images are suggested on image notices but performance of hub is not stable most of the time, and in general, it adds around 45-120 secs to build times depend on time and probably the load of Docker's hub service.

jnrbarros (approved):

haroonc:
Closing this as node versions are updated by PRs #973 and #978

PR #724, opened 2020-10-05
Add appengine example for testing

loosebazooka:
@bendory 
So there's some other issue, which i'm not sure how to fix: local build and run fails on a fresh repo because the `gcr.io/<project>/mvn:latest` latest isn't available yet when running the appengine/gcloud build?

### DIFF
--- a/mvn/cloudbuild.yaml
+++ b/mvn/cloudbuild.yaml
@@ -60,6 +60,9 @@ steps:
 - name: 'gcr.io/$PROJECT_ID/mvn:gcloud'
   args: ['install']
   dir: 'examples/spring_boot'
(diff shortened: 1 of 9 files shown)


mvn/cloudbuild.yaml
```diff
@@ -60,6 +60,9 @@ steps:
 - name: 'gcr.io/$PROJECT_ID/mvn:gcloud'
   args: ['install']
   dir: 'examples/spring_boot'
+- name: 'gcr.io/$PROJECT_ID/mvn:appengine'
+  args: ['package', 'appengine:stage']
```
loosebazooka, line 64: this runs a command that doesn't call out to a remote service but still requires the cloud sdk

bendory:
LGTM; just running a build before mergine.

Issue #725, opened 2020-10-06
[BUG] gke-deploy / PodDisruptionBudget


feitnomore:
## Failing to use PodDisruptionBudget
When using `PodDisruptionBudget` with `spec: maxUnavailable`, although resource is being deployed, the builder can't identify it as ready/healthy.

## Affected builder image
`gcr.io/cloud-builders/gke-deploy`

## Expected Behavior
Trigger should pass however builder fails do identify the `PodDisruptionBudget` is working fine.

## Actual Behavior
According to `resource/ready.go` under `podDisruptionBudgetIsReady`, one of the things that is being checked is `status.desiredHealthy == spec.minAvailable`, however if we deploy something using `maxUnavailable` instead of `minAvailabe`, that is supported on Kubernetes 1.17+, builder will always return false here:

```
minAvailable, ok, err := unstructured.NestedInt64(obj.Object, "spec", "minAvailable")
if err != nil {
	return false, fmt.Errorf("failed to get spec.minAvailable field: %v", err)
}
if !ok {
	return false, nil
}
```

## Steps to Reproduce the Problem

Use `maxUnavailable` instead of `minAvailable`.

```
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: nodejs04
  namespace: default
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app: nodejs04
```
## Additional Info
Suggestion here is to remove `status.desiredHealthy == spec.minAvailable` and `minAvailable` retrieval, and leave only checking is `status.currentHealthy >= status.desiredHealthy` as I think it should be sufficient.

If you are ok with the suggestion, I can try to submit a PR myself.

Thanks

feitnomore:
PR merged into master.

PR #726, opened 2020-10-06
Removing minAvailable from podDisruptionBudgetIsReady

feitnomore:
This PR refers to #725 
When using PodDisruptionBudget with spec: maxUnavailable, although resource is being deployed, the builder can't identify it is ready/healthy as it is currently checking for spec: minAvailable. So unless we have the minAvailable field in our object, it'll fail. 

What was done was remove the original `status.desiredHealthy == spec.minAvailable` check.


joonlim:
LGTM

joonlim:
@bendory Please merge

Issue #734, opened 2020-11-06
Add pyopenssl to gsutil builder


MattiaPance:
Hi all,
I want to use the _gsutil signurl_ command within a Cloud Build step with the gsutil builder, but when I tried it, the builder failed with:

'_CommandException: The signurl command requires the pyopenssl library (try pip install pyopenssl or easy_install pyopenssl)_'

Is it possible to fix it by adding pyopenssl to the gsutil builder?

feitnomore:
I've tried adding pyopenssl to the main package, but starts to throw some CryptographyDeprecationWarnings.
Not sure how to proceed here.

bendory:
Q: does `gsutil signurl` work with these images? https://github.com/GoogleCloudPlatform/cloud-sdk-docker

feitnomore:
doesn't looks like:

```
$ sudo docker run -it gcr.io/google.com/cloudsdktool/cloud-sdk gsutil signurl key.json gs://pizza-pizza-pizza/parisim.png
CommandException: The signurl command requires the pyopenssl library (try pip install pyopenssl or easy_install pyopenssl)
```

bendory:
I'm closing this issue here; please open a new issue over at https://github.com/GoogleCloudPlatform/cloud-sdk-docker which is owned by the Cloud SDK team. This appears to be an issue with `gsutil` or with its associated packaging with the python runtime in these Docker iamges.

bendory:
(Note: I would have just transferred the issue over to that repo, but I don't have permission.)

PR #735, opened 2020-11-06
Add signurl support to gsutil

feitnomore:
This is a commit to add pyopenssl to gsutil cloud builder for signurl support.

[Ticket #734]


feitnomore:
Doing some tests it still complains about pyopenssl although its installed. Gonna troubleshoot it better.

PR #756, opened 2021-02-23
Updated README and added comments around modules.

bendory:


go/prepare_workspace.inc
```diff
@@ -13,10 +13,17 @@
 # limitations under the License.
 
 prepare_workspace() {
+	# Use Go modules (https://golang.org/ref/mod) if defined. This is the
+	# prefered workspace setup; it has been available since Go 1.11 and is the
+	# default behavior as of 1.16.
```
ronanww, line 18: indentation seems off. It looks like here it was done with `tab` but in the rest of the file the indentation was done with `spaces`

go/prepare_workspace.inc
```diff
@@ -13,10 +13,17 @@
...
     if [[ -e "go.mod" ]]; then
         return 0
     fi
 
+	# Workspace setup without modules lives on for legacy purposes; it is no
+	# longer a supported behavior. Users should upgrade to use Go modules.
+	# See https://golang.org/ref/mod
+
```
ronanww, line 26: same comment as above ^

go/prepare_workspace.inc
bendory: thanks, fixed

go/prepare_workspace.inc
bendory: thanks, fixed

ronanww (approved):
👍

Issue #759, opened 2021-03-08
[BUG] Go builds with set GIT_CONFIG_VAL broken after #755


mlesar:
## Affected builder image

All images build after #755 merge.

## Expected Behavior

If `GIT_CONFIG_VAL` is set it should be used for fetching from private repositories.

## Actual Behavior

Currently, you get an error:
>fatal: could not read Username for 'https://github.com': terminal prompts disabled

mlesar:
In our case, we use `GIT_CONFIG_VAL` to add the GitHub access token to the GitHub URL so that we can fetch data from private repositories.

Example:
```yaml
steps:
- name: gcr.io/cloud-builders/go:latest
  id: go-build
  args: ['build', '-o', 'service', 'main.go']
  env:
    - 'GIT_CONFIG_VAL=url."https://${_GITHUB_ACCESS_TOKEN}@github.com/".insteadOf "https://github.com/"'
    - 'GOPRIVATE=github.com/XXXX'
...
```

PR #760, opened 2021-03-08
Moving git config above the go.mod check

mlesar:
#759

### DIFF
--- a/go/prepare_workspace.inc
+++ b/go/prepare_workspace.inc
@@ -13,6 +13,11 @@
 # limitations under the License.
 
 prepare_workspace() {
+    # If $GIT_CONFIG_VAL is set, set global git config value
+    if [[ "$GIT_CONFIG_VAL" ]]; then
+        eval "git config --global $GIT_CONFIG_VAL"
+    fi
+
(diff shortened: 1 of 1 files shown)


bendory:
@mlesar This looks to me like a bug fix -- thank you!

Can you please provide a more extensive explanation of what the issue is and how this fixes it? Referencing an example would also be helpful. (A test case that fails before / passes with this PR would be the best example.)

Happy to see that explanation here in the PR, or equally happy to see you file a descriptive issue which is then fixed by this PR, whatever you prefer.

Will look to get this merged ASAP following the above.

Thank you!

bendory:
Oh -- doh! -- looks like you indeed did most of the above already, I clearly should have noticed the attached issue. Would you mind please adding an example per my prior comment? (Either here or in the attached issue is fine.)

Thank you again!

mlesar:
> Would you mind please adding an example per my prior comment?

Added an example to the issue, let me know if anything else is needed.

bendory (approved):

bendory:
Thanks!

PR #767, opened 2021-03-24
Assuming "0" when `status` fields are missing from a deployment

### LINKED ISSUE #766: [BUG] when setting a deployment replicas to zero, build times out
## Affected builder image

gcr.io/cloud-builders/gke-deploy

## Expected Behavior

When scaling down a deployment (setting its `replicas` to `0`), gke-deploy should call `kubectl apply`, confirm that the number of replicas has changed, and report success.

## Actual Behavior

When scaling down a deployment (setting its `replicas` to `0`), gke-deploy keeps on calling `kubectl get Deployment POD NAME -n NAMESPACE --output=yaml`, until it times out.
It's like its success state depended on having a running deployment, although the goal was indeed to scale it down

1. set up a cloud build pipeline, have it spin up a container with multiple replicas
2. change the replicas to zero, submit the change to cloudbuild, check how it keeps on "polling" the deployment until it times out

ffissore:
Apparently, if a deployment has zero replicas, kubernetes yaml output will **not** contain fields `replicas`, `readyReplicas`, and `availableReplicas`

Function `deploymentIsReady` was instead assuming their presence, thus making `gke-deploy run` fail with a timeout when deploying a deployment downscale

With this PR, absence of the field is treated as if the field had value `0`, and made our deployment succeed

Fixes #766

### DIFF
--- a/gke-deploy/core/resource/ready.go
+++ b/gke-deploy/core/resource/ready.go
@@ -121,27 +121,27 @@ func deploymentIsReady(ctx context.Context, obj *Object) (bool, error) {
 		return false, nil
 	}
 
-	statusReplicas, ok, err := unstructured.NestedInt64(obj.Object, "status", "replicas")
+	statusReplicas, _, err := unstructured.NestedInt64(obj.Object, "status", "replicas")
 	if err != nil {
 		return false, fmt.Errorf("failed to get status.replicas field: %v", err)
 	}
-	if !ok || statusReplicas != specReplicas {
+	if statusReplicas != specReplicas {
 		return false, nil
 	}
 
-	readyReplicas, ok, err := unstructured.NestedInt64(obj.Object, "status", "readyReplicas")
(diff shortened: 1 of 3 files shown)


joonlim:
Could you add a test/tests for this?

ffissore:
Added

joonlim (approved):

ffissore:
@joonlim has this change been released? We keep on having the same issue on our builds

joonlim:
The latest changes should be available. Just making sure - are you using the `gcr.io/cloud-builders/gke-deploy` container?

joonlim:
@ffissore Apologies, after some digging we found that the changes are not rolled out yet. We are working on this ASAP.

ffissore:
Thank you @joonlim, can't wait to see this shipped

ffissore:
hey, @joonlim we are still experiencing the issue. can you confirm the image has been shipped? in our cloudbuild files we just have `- name: 'gcr.io/cloud-builders/gke-deploy'` which I assume means `latest`

PR #784, opened 2021-06-30
go: update go dockerfiles to use go1.16

cpanato:
go: update go dockerfiles to use go1.16

### DIFF
--- a/go/Dockerfile.alpine
+++ b/go/Dockerfile.alpine
@@ -1,4 +1,4 @@
-FROM golang:1.15-alpine
+FROM golang:1.16-alpine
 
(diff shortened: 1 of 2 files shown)


ronanww:
hi @cpanato, in https://github.com/GoogleCloudPlatform/cloud-builders/pull/786 we made the Go 1.16 builders available. You can now use `gcr.io/cloud-builders/go:1.16`

As of right now we need to keep the default builder to be `1.15` due to backwards compatibility as specified in https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/go#go-modules. Please let us know if using the Go1.16 builder helps you for your case

cpanato:
thanks @ronanww !!

PR #21, opened 2017-03-20
Add wrapper scripts to PATH, so that the entrypoints can be simple commands.

skelterjohn:

PR #789, opened 2021-07-22
Py2 has not been required to run the installer for a couple years now.

willfaris:
You don't need to use python 2 to run install.sh anymore


bendory (commented):
LGTM

Can you clarify please:
- is line 21 still needed?
- any changes needed on line 4?

willfaris:
regarding #4 and #21, I don't think you need any python 2 for gcloud but I was too risk averse to yank those lines, i don't know the full use case of this image (looks like the entry point just runs gcloud but I don't know the whole story).  I did confirm that gcloud is picking up Python 3.5 (preferring Py3 over Py2 which has been happening for over a year now.  You could probably make a smaller image by not installing any python 2 (nor the py2 crcmod), I did not feel comfortable doing that as you never know what downstream things are picking up this image and depending on things.  But from a gcloud perspective Python 2 does not need to be present.

bendory:
SGTM, closing this discussion, thanks for your help!

PR #791, opened 2021-07-29
Fix gke-deploy run apply with GCS input to point to use "*" to copy just the files and not the directory itself to allow apply to work without passing the recursive flag.

joonlim:


gke-deploy/core/gcs/gcs.go
```diff
@@ -28,17 +27,14 @@ type GCS struct {
 	GcsService services.GcsService
 }
 
-// Download copies file(s) from GCS. <dst> should be a directory, not a path to a file
+// Download copies file(s) from GCS. <dst> should be a directory, not a path to a file.
 func (s *GCS) Download(ctx context.Context, src, dst string, recursive bool) error {
-	log.Printf("Downloading file(s) %s from %s.\n", src, dst)
```
ChrisGe4, line 33: Why delete the log?

ChrisGe4 (approved):

gke-deploy/core/gcs/gcs.go
joonlim: I'm trying to keep this file more consistent with the rest of the code, where we don't print too much of what's happening under the hood unless -V is passed.

PR #795, opened 2021-09-07
add node16 and update lts, latest, and current tagged images

polyomino-24:
I added Node.js v16, which is latest and current version according to [docker hub](https://hub.docker.com/_/node), and moved LTS tag to v14.

### DIFF
--- a/npm/cloudbuild.yaml
+++ b/npm/cloudbuild.yaml
@@ -38,17 +38,20 @@ steps:
   - 'build'
   - '--build-arg=NODE_VERSION=12.18.3'
   - '--tag=gcr.io/$PROJECT_ID/npm:node-12.18.3'
-  # 12.18.3 is tagged :lts
-  - '--tag=gcr.io/$PROJECT_ID/npm:lts'
-  - '--tag=gcr.io/$PROJECT_ID/nodejs/npm'
   - '.'
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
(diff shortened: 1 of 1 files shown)


npm/cloudbuild.yaml
```diff
@@ -65,10 +68,12 @@ steps:
...
 - name: 'gcr.io/$PROJECT_ID/npm:node-12.18.3'
   args: ['--version']
-- name: 'gcr.io/$PROJECT_ID/npm:node-14.10.0'
+- name: 'gcr.io/$PROJECT_ID/npm:node-14.17.6'
+  args: ['--version']
+- name: 'gcr.io/$PROJECT_ID/npm:node-16.8.0'
   args: ['--version']
 
-# Test the examples with :latest
+# Test the examples with :latest and :current
```
polyomino-24, line 76: I think it would be better to test in LTS version too. What do you think?

jordanebelanger:
> @polyomino-24 Now that Node 16 is LTS, it would be great if we could (1) get the node 16 image in to the registry, and (2) move the LTS tag accordingly.

++ for this, even the documentation on GCP specifies that node 16 is now LTS, come on people

marcushultman:
@jordanebelanger The Node community maintains [official Node.js docker images](https://hub.docker.com/_/node?tab=description), which is likely why Google is not maintaining the same images anymore.

arvinddayal:
#869 should take care of this.

PR #800, opened 2021-10-01
Add gcloud*:ubuntu1804 builders and update git/curl builders

leeonlee:
Updating the git and curl builders updates the openssl versions which is
required for new Let's Encrypt certs

### DIFF
--- a/curl/Dockerfile
+++ b/curl/Dockerfile
@@ -1,3 +1,3 @@
-FROM launcher.gcr.io/google/ubuntu1604
+FROM gcr.io/gcp-runtimes/ubuntu_18_0_4
 COPY notice.sh /usr/bin
 ENTRYPOINT ["/usr/bin/notice.sh"]
--- a/gcloud/Dockerfile.slim.ubuntu1804
+++ b/gcloud/Dockerfile.slim.ubuntu1804
@@ -1,0 +1,34 @@
+FROM gcr.io/gcp-runtimes/ubuntu_18_0_4
+
+RUN apt-get -y update && \
+    apt-get -y install gcc python2.7 python-dev python-pip python3-pip wget ca-certificates \
+       # These are necessary for add-apt-respository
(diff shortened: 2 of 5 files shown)


georgecrawford:
@leeonlee @gbjk Hoping you can help here: 

How can I ensure that my Cloud Build jobs actually _use_ these new images? If I specify `gcr.io/cloud-builders/git@sha256:8e856e02dedddeebb49f68087623e085f05f3a8b0e07c4cf71c21efb5c181513` as the `name` parameter, I can clone fine from a repo behind Let's Encrypt, but if I use `gcr.io/cloud-builders/git:latest` or `gcr.io/cloud-builders/git` I'm still getting cert errors. 

Is there a document or repo somewhere where I can understand when the servers which run the builds will receive a new cache of builder images? Or is there a way I can speed that up without having to endure the 3+ minutes to pull the specific SHA for a builder which I know works? 


Thanks for any advice!

leeonlee:
@georgecrawford  Apologies for the situation. 

The changes will take until the end of the week to fully roll out. For the time being we also added a new cached tag `gcr.io/cloud-builders/git:openssl-upgrade` that contains the changes; this tag should be available today but don't expect this tag to be updated in the future.

I thought there would be some communication regarding the tag but I'll take a look.

@gbjk will update the docs, thanks.

georgecrawford:
Thanks Leon. That’s extremely helpful. If there’s a page thay I should visit to see updates and comms like this, or somewhere which lists the versions of the cached images, please you could let me know?


________________



George

georgecrawford:
@leeonlee Are you able to say if this is rolled-out yet as the new default? I'm still seeing cert errors when using the image `gcr.io/cloud-builders/git` (which I'd prefer to use over the `openssl-upgrade` tag, as I don't want to have to make a further round of changes to all our projects!). 

Many thanks

PR #807, opened 2021-10-15
Finish updating images that use ubuntu 1604 to 1804

leeonlee:
Ubuntu 1604 is EOL and is no longer receiving security updates.


leeonlee:
Forgot to hit amend..

PR #808, opened 2021-10-15
Update images that use ubuntu 16.04 to 20.04

leeonlee:
Ubuntu 1604 is EOL and is no longer receiving security updates.


dinvlad:
Would it be possible to update images inside Cloud Build also?

PR #836, opened 2022-05-23
Simplify Docker builds of multiple versions.

bendory:
- pass DOCKER_VERSION as arg to `docker build`
- versions from the same base image to share layers in both the local build and in the repository
- removed rebuild using `:latest`, which didn't test anything not already tested
- install from `stable` channel only
- fixed tags for `20.10` builds -- we build `20.10.14`, not `.13`


docker/cloudbuild.yaml
```diff
@@ -2,37 +2,58 @@
...
   id: '19.03.9'
+  wait_for: ['base']
 
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
   - '--tag=gcr.io/$PROJECT_ID/docker:20.10.13'
-  - '--file=Dockerfile-20.10.13'
+  - '--file=Dockerfile-versioned'
+  - '--build-arg=DOCKER_VERSION=5:20.10.14~3-0~ubuntu-focal'
```
leeonlee, line 33: We're building 20.10.14 then tagging it 20.10.13.

docker/cloudbuild.yaml
```diff
@@ -2,37 +2,58 @@
...
+  id: 'base'
+
 # Build all supported versions.
 - name: 'gcr.io/cloud-builders/docker'
   args:
   - 'build'
   - '--tag=gcr.io/$PROJECT_ID/docker:19.03.9'
-  - '--file=Dockerfile-19.03.9'
+  - '--file=Dockerfile-versioned'
+  - '--build-arg=DOCKER_VERSION=5:19.03.9~3-0~ubuntu-focal'
```
leeonlee, line 22: Can we use substitutions to reduce repetition for DOCKER_VERSION?

docker/cloudbuild.yaml
```diff
@@ -2,37 +2,58 @@
...
 # Future supported versions of docker builder go here.
 
 # Test each version by using it to run "docker build" on itself.
 - name: 'gcr.io/$PROJECT_ID/docker:19.03.9'
   args:
   - 'build'
-  - '--file=Dockerfile-19.03.9'
+  - '--file=Dockerfile-versioned'
+  - '--build-arg=DOCKER_VERSION=5:19.03.9~3-0~ubuntu-focal'
+  - '--build-arg=RELEASE_CHANNEL=edge'
```
leeonlee, line 47: What's the reason we're using edge for 19 and stable for 20?

docker/cloudbuild.yaml
bendory: Good catch; fixed.

docker/cloudbuild.yaml
bendory: Nice idea, done.

docker/cloudbuild.yaml
bendory: I'm guessing it's historical -- we presumably at one point in history installed a version that was not available on `stable` and since then it's just persisted.

Both current versions are available on `stable`, so removed, thanks for asking about thi.

Issue #835, opened 2022-05-23
[FR] Add `docker compose` to the official `gcr.io/cloud-builders/docker` image


akauppi:
## Detailed Feature Information

**Which builder is relevant to this request?**

`gcr.io/cloud-builders/docker:20.10.13`

**Is your feature request related to a problem? Please describe.**

I'm using Docker Compose in Cloud Build scripts, and currently need to bring in an external image ([`docker/compose`](https://hub.docker.com/r/docker/compose)). This image is stuck at 1.29.2 since a year - likely due to Docker moving from `docker-compose` CLI to `docker compose` (a plugin to the main `docker` command).

This plugin is not installed on the Cloud Build image, as of now.

**Describe the solution you'd like**

Adding `apt-get install docker-compose-plugin` (instructions [here](https://docs.docker.com/compose/install/)) to the Cloud Build Docker image.

I tried this. It works. I don't know about the size implications to the image though.

**Describe alternatives you've considered**

I'm currently using the `docker/compose` image. It works, but being stale I wish to have an upgrade path.

Considering making my own Docker image that is based on `gcr.io/cloud-builders/docker`. With the hopes that pulling such would be faster in Cloud Build than pulling a completely external image. Have not tried. 

**Additional context**

Thanks for considering.

PR #837, opened 2022-05-23
Install docker-compose packages.

bendory:
Fixes https://github.com/GoogleCloudPlatform/cloud-builders/issues/835

### DIFF
--- a/docker/Dockerfile-base
+++ b/docker/Dockerfile-base
@@ -11,4 +11,4 @@ RUN apt-get -y update && \
     apt-key fingerprint 0EBFCD88 && \
     add-apt-repository \
        "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
-	apt-get clean
+    apt-get clean
--- a/docker/Dockerfile-base
+++ b/docker/Dockerfile-base
@@ -8,4 +8,7 @@ RUN apt-get -y update && \
         make \
         software-properties-common && \
     curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
-    apt-key fingerprint 0EBFCD88
+    apt-key fingerprint 0EBFCD88 && \
+    add-apt-repository \
+       "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
+	apt-get clean
--- a/docker/Dockerfile-versioned
+++ b/docker/Dockerfile-versioned
@@ -2,9 +2,8 @@ FROM temp
 
 ARG DOCKER_VERSION=VERSION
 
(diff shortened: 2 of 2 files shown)


docker/Dockerfile-base
```diff
@@ -8,4 +8,7 @@ RUN apt-get -y update && \
         make \
         software-properties-common && \
     curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
-    apt-key fingerprint 0EBFCD88
+    apt-key fingerprint 0EBFCD88 && \
+    add-apt-repository \
+       "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
+	apt-get clean
```
leeonlee, line 14: Dedent

docker/Dockerfile-versioned
```diff
@@ -2,9 +2,8 @@ FROM temp
 
 ARG DOCKER_VERSION=VERSION
 
-RUN add-apt-repository \
-       "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
-    apt-get -y update && \
-    apt-get -y install docker-ce=${DOCKER_VERSION} docker-ce-cli=${DOCKER_VERSION}
+RUN apt-get -y install docker-ce=${DOCKER_VERSION} docker-ce-cli=${DOCKER_VERSION} \
+	    docker-compose docker-compose-plugin &&\
+	apt-get clean
```
leeonlee, line 7: Also dedent

leeonlee, line 7: Also dedent

docker/Dockerfile-versioned
```diff
@@ -2,9 +2,8 @@ FROM temp
 
 ARG DOCKER_VERSION=VERSION
 
-RUN add-apt-repository \
-       "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" && \
-    apt-get -y update && \
```
leeonlee, line 7: Do we need to run `apt-get update` after `add-apt-repository`? If so, we need this line

leeonlee, line 7: Do we need to run `apt-get update` after `add-apt-repository`? If so, we need this line

leeonlee (changes requested):

docker/Dockerfile-versioned
bendory: tab weirdness; cleaned up

docker/Dockerfile-versioned
bendory: apparently not -- everything works without it (which admittedly surprised me after adding the repository)

PR #841, opened 2022-06-30
node: adding node lts version 16.15.1 & updating tags

ramthir:
adding node lts version 16.15.1 & updating tags as we need them for the latest versions of angular and other libraries.


ramthir:
Closing this as #842 address this.

PR #848, opened 2022-08-05
updated to --short

sanketbisne:
We wants our commits more shorts in our pipelines , so changing the command to --short keyword.


haroonc:
Closing as author has not signed Contributor License Agreement (CLA)

PR #861, opened 2022-09-27
Fix bug in Dockerfile of Twine Cloud Builder

amcooper:
Replace `CMD` with `ENTRYPOINT` in the Dockerfile.


leeonlee (approved):
nit: please use a more descriptive commit title

PR #898, opened 2023-03-02
Update default go image to 1.20

hazim1093:
Update default go image to the latest (1.20)


arvinddayal:
Sorry for the delay. This looks fine but would you mind re-organizing cloudbuild.yaml so that v1.20 is up top (where v1.15 currently is) and moving 1.15 down? Thanks!

PR #905, opened 2023-04-13
[mvn] pin :latest tag to version 3.8 and add 3.9 tag

mwiczer:
Pinning :latest to version 3.8 since there's a decent selection of [breaking changes](https://maven.apache.org/docs/3.9.0/release-notes.html#potentially-breaking-core-changes-if-migrating-from-3-8-x) from 3.8.x to 3.9.

Also add :3.9.1 tag to the builder in case people would like to upgrade their builder.

### DIFF
--- a/mvn/cloudbuild.yaml
+++ b/mvn/cloudbuild.yaml
@@ -36,11 +36,23 @@ steps:
   - '.'
   waitFor: ['-']
   id: '3.8'
-
-# Build the latest version.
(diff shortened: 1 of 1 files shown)


GiangN (approved):

mmartinekTX:
Bug opened for this change: https://github.com/GoogleCloudPlatform/cloud-builders/issues/908

PR #921, opened 2023-05-29
Create a new image for bazel 6.2.0

### LINKED ISSUE #849: [FR] Update Bazel Version
Latest images are bazel 3.5.0, yet 5.2.0 is the current release

### LINKED ISSUE #884: [FR] Update bazel to 6.0.0
## Detailed Feature Information

**Which builder is relevant to this request?**
gcr.io/cloud-builders/bazel

**Is your feature request related to a problem? Please describe.**
Per @khalkie's comment on #880, 6.0.0 is the latest stable version of bazel. 

**Describe the solution you'd like**
Potential options in order of personal preference:

1. Unpin the version. It's reasonable to count on customers with production-critical bazel dependencies to lock to a specific version (ex. I have one project that depends on `latest` and a more sensitive project that specifies a specific hash).
2. Unpin the version, and tag the uploaded version going forward. 
3. Create a gcr.io/cloud-builders/bazel-6 builder.

SpencerC:
Fixes #884
Fixes #849

### DIFF
--- a/bazel-6/.gcloudignore
+++ b/bazel-6/.gcloudignore
@@ -1,0 +1,3 @@
+cloudbuild.yaml
+.gcloudignore
+README.md
--- a/bazel-6/Dockerfile
+++ b/bazel-6/Dockerfile
@@ -1,0 +1,58 @@
+FROM gcr.io/gcp-runtimes/ubuntu_20_0_4
+
+ADD bazel.sh /builder/bazel.sh
+ARG DOCKER_VERSION=5:19.03.9~3-0~ubuntu-focal
+
+RUN \
+    # This makes add-apt-repository available.
+    apt-get update && \
+    apt-get -y install \
+        python \
+        python3 \
+        python-pkg-resources \
+        python3-pkg-resources \
+        software-properties-common \
+        unzip && \
+
+    # Install Git >2.0.1
+    add-apt-repository ppa:git-core/ppa && \
+    apt-get -y update && \
+    apt-get -y install git && \
+
+    # Install bazel (https://docs.bazel.build/versions/master/install-ubuntu.html)
+    apt-get -y install openjdk-8-jdk && \
+    echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list && \
+    curl https://bazel.build/bazel-release.pub.gpg | apt-key add - && \
+    apt-get update && \
+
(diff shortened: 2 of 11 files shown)


hulkholden:
@SpencerC are you still working on this? I am also in need of bazel 6 so I've been putting a PR together based the approach @arvinddayal suggests. Let me know if you think I should keep going.

One thing I noticed is that the existing example is incompatible with bazel 6. I'm not sure if it's worth maintaining separate tests for 5.x and 6.x, or just testing the most latest version (which is what the npm builder does).

SpencerC:
@hulkholden I probably won't get back around to this for at least another week, so if you want to keep working on yours/join our efforts that works for me.

The changes I made in https://github.com/GoogleCloudPlatform/cloud-builders/pull/921/commits/bde4c93a7b5b90c1173e61eee28e13b13eb6c732 modify the tests to work with Bazel 6. I wouldn't mind if you leveraged them in a separate PR if you don't mind listing me as a [co-author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)🙂.

hulkholden:
Thanks @SpencerC - with your fixes I can get the tests to pass with bazel 6 - WIP at https://github.com/hulkholden/cloud-builders/tree/bazel6. Can I use SpencerC@users.noreply.github.com for the coauthor or is it id+username@ for your account?

@arvinddayal One concern I have with the approach you suggest is that anyone who hasn't already pinned to a specific hash is potentially going to be broken as soon as they start picking up bazel 6 rather than bazel 5. I suspect this will be a lot of people.

The workaround will be for them to explicitly pin to 5.4.0 but it won't be particularly obvious why they are suddenly broken and how they should go about fixing it (it wasn't obvious to me when the example/ tests broke because the errors are deep in imported modules like rules_go).

What's your general approach to these kinds of breaking changes? Perhaps we could point :latest at 5.4.0 initially which would give people time to migrate?

SpencerC:
@hulkholden last time they just used my email: https://github.com/mit-dci/opencbdc-tx/commit/f617c4e8eace2d0d76974cc3b3fac41de85c089c

hulkholden:
Thanks @SpencerC - PTAL https://github.com/GoogleCloudPlatform/cloud-builders/pull/925 and check that I attributed you correctly :)

I realised we can explicitly tag the 5.4.0 version with :latest to avoid breaking people. We probably still want to move that to :6.x at some point but at least now people can start explicitly pinning to the versions they need.

PR #935, opened 2023-08-25
Future supported: Docker 24 tagged versions

markmandel:
I have a strong sense that I'm going to need features in Docker 24 shortly, so wanted to help get it into at least a tagged version (not the default of course!) so it was possible to take them for a spin on Cloud Build.

### DIFF
--- a/docker/cloudbuild.yaml
+++ b/docker/cloudbuild.yaml
@@ -4,6 +4,7 @@
 substitutions:
   _DOCKER_19: DOCKER_VERSION=5:19.03.9~3-0~ubuntu-focal
   _DOCKER_20: DOCKER_VERSION=5:20.10.14~3-0~ubuntu-focal
+  _DOCKER_24: DOCKER_VERSION=5:24.0.6-1~ubuntu.20.04~focal
 
 steps:
 # Pre-requisite: build the base OS container; 'temp' is referenced in the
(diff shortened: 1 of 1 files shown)


markmandel:
Gentle bump - seeing if we can push this through?

docker/cloudbuild.yaml
```diff
@@ -41,7 +42,17 @@ steps:
...
+  args:
+    - 'build'
+    - '--tag=gcr.io/$PROJECT_ID/docker:24.0.6'
+    - '--file=Dockerfile-versioned'
+    - '--build-arg=${_DOCKER_24}'
+    - '.'
+  id: '24.0.6'
+  wait_for: [ 'base' ]
+
+  # Test each version by using it to run "docker build" on itself.
```
leeonlee, line 55: Can you move this back to line 44?

leeonlee (dismissed):
Please remove 23 from the commit message since it's been removed

docker/cloudbuild.yaml
markmandel: That should look better - PTAL!

docker/cloudbuild.yaml
```diff
@@ -58,8 +59,28 @@ steps:
...
+    - '--build-arg=${_DOCKER_24}'
+    - '.'
+  id: '24.0.6'
+  wait_for: [ 'base' ]
+
+
+
 # Tests for future supported versions of docker builder go here.
 
+- name: 'gcr.io/$PROJECT_ID/docker:24.0.6'
```
markmandel, line 76: @leeonlee did you want this pulled up above this header as well?

leeonlee: It would be more consistent to move the step that builds and tags 24.0.6 (lines 62-70) under the "Build all supported versions." header (~line 42)

leeonlee (approved):

docker/cloudbuild.yaml
markmandel: Sorry - got all skewiff. That should all be good now.

leeonlee (approved):

PR #942, opened 2023-08-31
Parse generation from Storage URI

ronanww:


gcs-fetcher/pkg/common/common.go
```diff
@@ -42,22 +43,47 @@ type ManifestItem struct {
 //
```
arvinddayal, line 43: Should we include generation in the 2 URI examples above?

ronanww: Good idea, done

arvinddayal (approved):

PR #966, opened 2023-12-08
Bump nodejs LTS and current/latest version

ii64:
Add `20.10.0` as the new LTS, and  `21.4.0` as the new current/latest. Dropping LTS tag for `18.12.0`, and current/latest tag for `19.0.0`.

### DIFF
--- a/npm/cloudbuild.yaml
+++ b/npm/cloudbuild.yaml
@@ -55,8 +55,6 @@ steps:
   args:
   - 'build'
   - '--build-arg=NODE_VERSION=18.12.0'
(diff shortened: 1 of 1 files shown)


austinzhao-go:
Hi @ii64, thanks for raising this support version update.

as for backward compatible and not breaking current users' usages, think here good to add the `20.10.0` and `21.4.0` node versions while we should not move the `lts`, `latests`, and `current` tags.

codewithgiang (approved):

codewithgiang:
Can we merge this? I'm working bringing a node project up to LTS and I am unable to use cloud build to do so because the versions

PR #1001, opened 2024-05-30
Update error message when pulling from GCS if error may be related to VPC SC.

kennethye1:
Ref internal bug: b/305977141


kennethye1:
@haroonc @kmontg PTAL if this is ok. Thanks.

haroonc (approved):

PR #1011, opened 2024-09-18
Upgrade Ubuntu from 20.04 to 22.04

yaoxiaoqi:
Upgrade Ubuntu from 20.04 to 22.04 because the git curlOptResolve option doesn't work with Ubuntu 20.04.


haroonc (approved):

liggitt:
> the git curlOptResolve option doesn't work with Ubuntu 20.04.

I think this dropped the `curl` CLI from being available in the image, which is disrupting several build scripts relying on this image (and can be used in things like git hooks, fetching metadata to populate git config with the service account email, etc)

PR #1017, opened 2024-09-24
Upgrade base image to ubuntu 22 and install curl

yaoxiaoqi:
Upgrade base image for gcloud to ubuntu 22 and install curl, openssl and bzip2


haroonc (approved):
Thanks @yaoxiaoqi for ensuring that the upgrade has all the tools that were available in Ubuntu 20.04 version of this image.

PR #1044, opened 2025-03-03
Modify pip install to pip install --require-hashes

Tiyas-13:


gcloud/Dockerfile.slim
```diff
@@ -19,8 +19,10 @@ RUN apt-get -y update && \
         --disable-installation-options && \
 
     # install crcmod: https://cloud.google.com/storage/docs/gsutil/addlhelp/CRC32CandInstallingcrcmod
-    pip install -U crcmod && \
-    pip3 install -U crcmod && \
+    COPY requirements.txt .
+    COPY requirements2.txt .
+    pip install --require-hashes -r requirements2.txt
```
edwardthiele, line 24: nit: consider naming the file something like requirements_py2 or something to make it clear

edwardthiele (approved):

gcloud/Dockerfile.slim
Tiyas-13: done

edwardthiele (approved):

PR #1101, opened 2026-05-24
docs: update Bazel public image guidance

### LINKED ISSUE #927: Docs refer to stale bazel-provided images, new images aren't compatible with Cloud Build
A couple of the READMEs refer to `gcr.io/cloud-marketplace-containers/google/bazel` as alternative images provided by the bazel team, but it looks like this repo hasn't been updated since 2020.

The [bazel docs](https://bazel.build/install/ubuntu) now point at `gcr.io/bazel-public/bazel` which does seem to be up-to-date.

However,  I was unable to get these images to work as a drop-in replacement with Cloud Build.

Small example:

```yaml
steps:
- id: 'Bazel version'
  name: '[gcr.io/bazel-public/bazel:6.2.1](http://gcr.io/bazel-public/bazel:6.2.1)'
  entrypoint: 'bazel'
  args: ['version']
```

Gives:

```
Step #0 - "Bazel version": Status: Downloaded newer image for [gcr.io/bazel-public/bazel:6.2.1](http://gcr.io/bazel-public/bazel:6.2.1)
Step #0 - "Bazel version": [gcr.io/bazel-public/bazel:6.2.1](http://gcr.io/bazel-public/bazel:6.2.1)
Step #0 - "Bazel version": FATAL: mkdir('/builder/home/.cache/bazel/_bazel_ubuntu'): (error: 13): Permission denied
Finished Step #0 - "Bazel version"
```

I think this is because the bazel [Dockerfile](https://github.com/bazelbuild/continuous-integration/blob/master/bazel/oci/Dockerfile) is setting up an `ubuntu` user rather than running as root. Another consequence of this is that `/workspace/` is not writeable which makes it painful to pass around outputs between build steps.

One workaround for the cache issue is to pass `--output_user_root` like this:

```
steps:
- id: 'Bazel version'
  name: '[gcr.io/bazel-public/bazel:6.2.1](http://gcr.io/bazel-public/bazel:6.2.1)'
  entrypoint: 'bazel'
  args: ['--output_user_root=/workspace/.bazel', 'version']
```

But that still doesn't resolve the `/workspace/` issues.

So how should the cloud-builders READMEs be updated? I can send a PR to update the stale reference to gcr.io/bazel-public/bazel, but as the images are not drop-in replacements I think it would cause more confusion.

Maybe https://github.com/GoogleCloudPlatform/cloud-builders/blob/master/bazel/README.md can provide a bit more detail and the top-level README could point there?

Maybe there's a deeper compatibility issue which needs resolving? It is surprising that two Google-maintained bits of build infrastructure don't play nicely together.

Haihan-Jiang:
## Description

Update the Bazel builder documentation to avoid pointing users at the stale `gcr.io/cloud-marketplace-containers/google/bazel` image.

- Point the top-level alternatives list and Bazel README at the current Bazel public image, `gcr.io/bazel-public/bazel`.
- Call out that the Bazel public image is not a direct drop-in replacement for `gcr.io/cloud-builders/bazel` in Cloud Build.
- Add a Cloud Build example that sets `entrypoint: bazel` and a writable `--output_user_root` for the official image's non-root `ubuntu` user.
- Note that cross-step outputs/cache should use `/workspace` or an explicitly writable Cloud Build volume.

Fixes #927

## Tests

- `git diff --check`
- Verified the current Bazel install docs reference `gcr.io/bazel-public/bazel` for Docker container usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ for the corresponding Cloud Builder.
   - [`openjdk`](https://hub.docker.com/_/openjdk) supports many production versions of Java across multiple platforms
   - [`gradle`](https://hub.docker.com/_/gradle/) supports a matrix of Java and gradle versions across multiple platforms
   - [`maven`](https://hub.docker.com/_/maven/) supports a matrix of Java and maven versions across multiple platforms
-- [`gcr.io/cloud-marketplace-containers/google/bazel`](http://gcr.io/cloud-marketplace-containers/google/bazel) is provided by the bazel team and runs the [`bazel`](https://bazel.build/) tool
+- [`gcr.io/bazel-public/bazel`](https://bazel.build/install/ubuntu#docker-container) is provided by the bazel team and runs the [`bazel`](https://bazel.build/) tool. It is not a direct replacement for `gcr.io/cloud-builders/bazel`; see [`bazel/README.md`](bazel/README.md) for Cloud Build usage notes.
 - `curl` is packaged in:
   - [`launcher.gcr.io/google/ubuntu1604`](https://console.cloud.google.com/launcher/details/google/ubuntu1604)
   - [`curlimages/curl`](https://hub.docker.com/r/curlimages/curl) is community-supported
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The change updates the Bazel public image guidance in the docs, and the visible checks are green.

PR #1102, opened 2026-05-24
gke-deploy: apply existing Namespace configs

### LINKED ISSUE #873: [BUG] gke-deploy does not update namespaces with new labels if the namespace already exists
## Notice

**If you are reporting an issue with the hosted Google Cloud Build service, please contact your Google Cloud Support representative or report your issue using [Google's public issue tracker](https://issuetracker.google.com/issues/new?component=190802&template=1162743).**

This GitHub issue tracker is intended for bugs with the officially supported builder images __only__. This issue tracker is __not__ the right place to report issues with triggers or notifications.

If you're not sure where to report your issue, use [Google's public issue tracker](https://issuetracker.google.com/issues/new?component=190802&template=1162743).

## Affected builder image

gcr.io/cloud-builders/gke-deploy

## Expected Behavior

Applying an existing namespace with a new label will update the labels to what is in the YAML

## Actual Behavior

Applying an existing namespace with a new label does not update the namespaces labels at all.

1. apply a namespace via a yaml file
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: dev
  labels:
    name: dev
    role: shared-ingress
```
2. `kubectl get namespaces --show-labels` shows it has the above labels
3 update that namespace to add a new label
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: dev
  labels:
    name: dev
    role: shared-ingress
    pod-security.kubernetes.io/audit: restricted
```
4. `kubectl get namespaces --show-labels` shows the labels have not been updated

## Additional Info
doing these same steps but instead using kubectl locally they all work.

Haihan-Jiang:
Fixes #873.

`gke-deploy apply` currently handles Namespace objects specially: it checks whether each Namespace already exists, creates it if missing, and then removes Namespace objects from the normal apply list. That means existing Namespace objects are never applied again, so changes such as new labels in the manifest are silently skipped.

This change keeps the existing create-warning behavior for missing namespaces, but always applies the Namespace manifest. Existing namespaces can then be updated by `kubectl apply`, matching the behavior users expect from applying the same YAML with kubectl directly. Namespace objects are still excluded from the deployment summary.

Validation:
- `go test ./deployer ./core/cluster ./core/resource` from `gke-deploy`
- `go test ./...` from `gke-deploy`
- `git diff --check`

### DIFF
--- a/gke-deploy/deployer/deployer.go
+++ b/gke-deploy/deployer/deployer.go
@@ -408,13 +408,13 @@ func (d *Deployer) Apply(ctx context.Context, clusterName, clusterLocation, clus
 			}
 			if !exists {
 				fmt.Fprintf(os.Stderr, "\nWARNING: It is recommended that namespaces be created by an administrator. Creating namespace %q because it does not exist.\n\n", nsName)
-				objString, err := resource.EncodeToYAMLString(obj)
-				if err != nil {
-					return fmt.Errorf("failed to encode obj to string")
-				}
-				if err := cluster.ApplyConfigFromString(ctx, objString, "", d.Clients.Kubectl); err != nil {
-					return fmt.Errorf("failed to apply Namespace configuration file with name %q to cluster: %v", nsName, err)
-				}
+			}
+			objString, err := resource.EncodeToYAMLString(obj)
+			if err != nil {
+				return fmt.Errorf("failed to encode obj to string")
+			}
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. Could someone take a look when convenient?

Issue #951, opened 2023-10-04
[FR] Add labels to cloud-builder images


haroonc:
"It would be neat if images built by GCB (v1 or v2) could populate OCI annotation keys as labels
https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys
" @kmontg

PR #1107, opened 2026-05-24
Add OCI labels to builder images

Haihan-Jiang:
## What changed

- Added OCI image labels to the Dockerfiles for the official builder images maintained in this repository.
- The labels set `org.opencontainers.image.source` to this repository and `org.opencontainers.image.vendor` to `Google LLC`.

## Why

This addresses #951 by making the published builder images expose common OCI annotation keys as Docker image labels.

## Validation

- `git diff --check`
- Verified every non-example top-level Dockerfile has both OCI labels.


Haihan-Jiang:
This is ready for review. It adds OCI image label metadata across the builder Dockerfiles, and the visible checks are green.

Issue #1056, opened 2025-04-18
[BUG]


howesda:
## Affected builder image

gcr.io/cloud-builders/mvn:appengine

## Expected Behavior

Build and deploy my GCP AppEngine Standard Edition Application via CloudBuild - this has been working correctly for many years

## Actual Behavior
Has started to fail, indicating that it can't find Python 

Downloaded from central: https://repo.maven.apache.org/maven2/com/google/guava/guava/32.1.2-jre/guava-32.1.2-jre.jar (3.0 MB at 8.2 MB/s)
Apr 18, 2025 6:59:04 AM com.google.cloud.tools.managedcloudsdk.install.Downloader download
INFO: Downloading https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz to /root/.cache/google-cloud-tools-java/managed-cloud-sdk/downloads/google-cloud-sdk.tar.gz
Welcome to the Google Cloud CLI!
WARNING: You appear to be running this script as root. This may cause 
the installation to be inaccessible to users other than the root user.
/root/.cache/google-cloud-tools-java/managed-cloud-sdk/LATEST/google-cloud-sdk/install.sh: line 216: python: command not found
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:51 min
[INFO] Finished at: 2025-04-18T06:59:09Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.google.cloud.tools:appengine-maven-plugin:2.5.0:deploy (default-cli) on project chattyappsdata: Execution default-cli of goal com.google.cloud.tools:appengine-maven-plugin:2.5.0:deploy failed: com.google.cloud.tools.managedcloudsdk.command.CommandExitException: Process failed with exit code: 127 -> [Help 1]

## Steps to Reproduce the Problem

I am using CloudBuilder with this yaml file

steps:
  - name: "gcr.io/cloud-builders/mvn:appengine"
    args: ["-D", "skipTests", "-D", "endpoints.project.id=$_ENDPOINTS_PROJECT_ID", "-D", "app.deploy.projectId=$_PROJECT_NAME","clean", "package", "appengine:deploy", "-P", "$_PROFILE"]
timeout: "1600s"

## Additional Info
This has been working successfully for many years now - my last successful build was 13/04/2025, 15:17 ( BST )

[...]

PR #1111, opened 2026-06-06
Fix: add python package to mvn:appengine builder

64johnlee:
## Summary

The `google-cloud-sdk` installer requires the `python` command to be available. The `Dockerfile.appengine` was installing `python3` but not the `python` package, which causes the SDK installation to fail with `python: command not found`.

This fix adds `python` to the apt-get install line, ensuring both `python` and `python3` are available for the Cloud SDK setup.

## Verification

- Fixes issue #1056 (mvn:appengine: Unable to find Python)
- Single-line change to Dockerfile.appengine  
- No side effects: `python` package in Debian/Ubuntu is just a symlink to python3
- Matches existing apt-get patterns in the Dockerfile

🤖 Generated with Claude Code

### DIFF
--- a/mvn/Dockerfile.appengine
+++ b/mvn/Dockerfile.appengine
@@ -3,7 +3,7 @@ ARG MAVEN_VERSION=latest
 FROM gcr.io/${PROJECT_ID}/mvn:${MAVEN_VERSION}
 
 # install python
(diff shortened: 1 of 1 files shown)


64johnlee:
gcla

